PR #48, opened 2014-12-16
Remove global `React` lookup during runtime

### LINKED ISSUE #47: [1.1.0-rc-1] React must be exposed globally for components to work.
I'm using the 1.1.0-rc-1 to try out the new components, and so far so good. However, I realized that React must be exposed globally (by the consumer of the library) in order for these components to work. Was that an intentional choice?

ericf:
This removes the global `React` lookup that was happening within the React Intl Components' `render()` methods. A new local `react` module has been created to provide the reference to the `React` object.

**Note:** This is a temp solution, the real fix we want is to update the ES6 Module Transpiler to poperly deal with external, non ES6 modules.

Closes #47

### DIFF
--- a/src/components/date.js
+++ b/src/components/date.js
@@ -1,5 +1,7 @@
-/* global React */
-/* jslint esnext:true */
+/* jshint esnext:true */
+
+// TODO: Use `import React from "react";` when external modules are supported.
+import React from '../react';
 
 import IntlMixin from '../mixin';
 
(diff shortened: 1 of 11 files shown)


deezahyn:
getting an error now

```
 Unexpected reserved word
You may need an appropriate loader to handle this file type.
| // supports external modules. This is a hack for now that provides the local
| // modules a referece to React.
| export default React;
| 
 @ ./~/react-intl/index.js 6:15-31
```

opened an issue
https://github.com/yahoo/react-intl/issues/124

PR #55, opened 2014-12-31
[Documentation] Adding contributing.md

jasonmit:
Lets iterate on this since I'm not entirely familiar with your process.

/cc @ericf

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,11 +1,14 @@
 # Contributing
 
+Pull requests are very welcome, but should be within the scope of the project, and follow the repository's code conventions. Before submitting a pull request, it's always good to file an issue, so we can discuss the details of the PR.
+
 ## Submitting a pull request
-1. Read the [contributor license agreement](https://yahoocla.herokuapp.com/)
+1. At Yahoo, we have a single [Yahoo Open Source Contributor License Agreement](https://yahoocla.herokuapp.com/) that we ask contributors to electronically sign before merging in their Pull Requests. Here's the CLA's human-readable summary:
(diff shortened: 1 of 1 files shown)


caridy:
I guess the info in formatjs.io is more about how to contribute to formatjs umbrella of projects, but the contributing.md seems necessary for github to provide meaningful information about how to do the right thing when submitting a PR, how to run tests, etc., I will be ok with a contributing.md per library, as we normally do for all other yahoo pkgs.

jasonmit:
:+1: let me know if the language should change at all or if this is a good enough starting point.  @ericf just expressed interest in needing a contributing.md a few weeks back

caridy:
wording is fine @jasonmit, I think we should add installation and test commands details, that's all from my side.

ericf:
My main concern is just to prompt people to look at the past issues before opening dups.

CONTRIBUTING.md
```diff
@@ -0,0 +1,31 @@
...
+5. Follow syntax guidelines detailed below.
+
+6. Push the changes to your fork and submit a pull request.  If this resolves any issues, please mark in the body `resolve #ID` within the body of your pull request.  This allows for github to automatically close the related issue once the pull request is merged.
+
+7. Last step, [submit the pull request][pr]!
+
+[pr]: https://github.com/yahoo/react-intl/compare/
+[fork]: https://github.com/yahoo/react-intl/fork/
+
+## Reporting a Bug
```
ericf: I'd move this above the PR section.

jasonmit:
@caridy I'll need to send a separate pull request to allow the tests to be run outside of saucelabs.

Issue #35, opened 2014-11-26
Examples of fallbacks?


pselden:
There is mention of support for providing fallbacks -- locales: ['fr-FR', 'en'], but I'm not at all clear on how you'd actually use it.

In the one example you show of using multiple languages for messages, you're explicitly passing in messages['en-US'] to the root component messages prop, so I don't see how you could ever have fallback logic when you're explicitly not passing in the messages of the other locales.

mschipperheyn:
@ericf I also don't think providing empty strings is the way to go. However, in the Java world the accepted and imho best solution is to have a flag that allows you to configure message processing so that in stead of an exception, the requested key is returned (e.g. 'ui.button.save'). Having an application fail completely just because of a missing text imho is ridiculous.

Dakuan:
@mschipperheyn Agree completely, we are going to have to run a fork for now, see #56

caridy:
@Dakuan, @mschipperheyn, I like the example of the "Save" button from @ericf, and a console.log just does not make the cut for us. The reality is simple, if you throw, and you provide a way for people to prevent that by implementing something custom, everyone has the ability to do what they want. But if we warn instead, what about those who want to enforce very strict rules, like we do at Yahoo?

Dakuan:
@caridy I'd suggest that the permissive behaviour should be the default (to make the lib friendly to work with) and for the strict behaviour to be an option or override...

Dakuan:
It really depends if you want something aligned with how you guys work internally or something that's going to be friendly enough for us mortals to use. I think it would be a shame if you went for the 1st option. 

Sorry to bang on! I really am extremely grateful that all this is available at all :bouquet:

pselden:
Not sure if it will be useful at all to this discussion, but here is what I came up with for providing fallbacks for each locale (uses gulp).

https://gist.github.com/pselden/093965d29fc95ec7026b

caridy:
@pselden that's a good starting point for a fallback mechanism. At yahoo, we also do fallback at the message level, saying: you might have `fr.json`, but a missing `SAVE` message within in, which is a new entry in `en.json`, and should be propagated, specially if you have a separate tool to consume `en.json` to produce all the other files based on those entries.

mschipperheyn:
[...]

PR #56, opened 2015-01-02
warn instead of throw

Dakuan:
Missing translation keys are unfortunately a fact of life - especially in development and when adding languages.

Throwing when keys are missing is a bit too severe for my liking. I'm proposing to log a warning, and return the key instead.

Potentially there's more work in both the logging, and returning something prettier than the raw key, but I thought I'd see what people thought of the principle first.

### DIFF
--- a/src/mixin.js
+++ b/src/mixin.js
@@ -118,9 +118,8 @@ export default {
                 return obj[pathPart];
             }, messages);
         } finally {
-            if (message === undefined) {
-                throw new ReferenceError('Could not find Intl message: ' + path);
-            }
+            console.warn('Could not find Intl message: ' + path);
+            return path;
         }
 
         return message;


yahoocla:

Dakuan:
@yahoocla done!

Dakuan:
ping?

caridy:
@Dakuan thanks for the PR, but we don't plan to change this behavior, instead, we plan to provide guidance, and even tools for people to do whatever is best for them. Part of the discussion is here: https://github.com/yahoo/react-intl/issues/35, including a couple of approaches.

Dakuan:
@caridy ok - hope there's something there soon, as it is we are having to run my fork, which always makes me nervous. Having read #35 I'm still very much of the opinion that throwing when there's a perfectly reasonable alternative course of action is bad in the general case, and doubly so in the specific case where the error is being thrown by an event thats likely to be frequent.

caridy:
I like the example of the "OK" button from @ericf, and a console.log just does not make the cut for us. The reality is simple, if you throw, and you provide a way for people to prevent that by implementing something custom, everyone has the ability to do what they want. But if we warn instead, what about those who want to enforce very strict rules, like we do at Yahoo?

PR #62, opened 2015-01-15
Remove number type check from formatNumber()

ericf:
This removes an unnecessary type check from `formatNumber()` since the underlying `Intl.NumberFormat#format()` method will never throw.

_Note:_ The point of the type check was to provide a more useful, higher-level error message to the users of React Intl so they know where the error is coming from. The date type checks are still useful since they are more descriptive than what the underlying `format()` method will throw.

### DIFF
--- a/src/mixin.js
+++ b/src/mixin.js
@@ -27,12 +27,6 @@ function assertIsDate(date, errMsg) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

caridy:
LGTM

PR #63, opened 2015-01-16
Merge Intl component options with named format

### LINKED ISSUE #61: Overriding/augmenting named format properties
Currently named formats are an either/or proposition -- if you use a named format you cannot pass additional format arguments to it to either add new format args or override named ones. It would be useful to be able to do this in cases where you know some of the specifiers you want at build time but others at run time.

Example:

```js
/*
{
  "formats": {
    "number": {
      "currency": {
        "style": "currency",
        "minimumFractionDigits": 2
      }
    }
  }
}
*/
var price = { value: 42000, currency: 'USD' } ; // assume this comes from some api and is not hard coded
  
return <IntlNumber format="currency" currency={price.currency}>{price.value}</IntlNumber>;
```

This will always render it using the current locale's currency specifier because the currency code is ignored.

ericf:
Fixes #61

### DIFF
--- a/src/mixin.js
+++ b/src/mixin.js
@@ -29,14 +29,14 @@ function assertIsDate(date, errMsg) {
 
 export default {
     statics: {
-        filterFormatOptions: function (obj) {
+        filterFormatOptions: function (obj, defaults) {
             return (this.formatOptions || []).reduce(function (opts, name) {
                 if (obj.hasOwnProperty(name)) {
                     opts[name] = obj[name];
                 }
 
                 return opts;
-            }, {});
+            }, defaults ? Object.create(defaults) : {});
         }
     },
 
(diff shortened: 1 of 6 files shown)


yahoocla:
CLA is valid!

src/mixin.js
```diff
@@ -29,14 +29,14 @@ function assertIsDate(date, errMsg) {
...
-        filterFormatOptions: function (obj) {
+        filterFormatOptions: function (obj, defaults) {
             return (this.formatOptions || []).reduce(function (opts, name) {
                 if (obj.hasOwnProperty(name)) {
                     opts[name] = obj[name];
                 }
 
                 return opts;
-            }, {});
+            }, defaults ? Object.create(defaults) : {});
```
ericf: Prototype inheritance is safe to use here because [ECMA-402's `GetOption` abstract operation performs a `[[Get]]`](http://www.ecma-international.org/ecma-402/1.0/index.html#sec-9.2.9).

tests/smoke/smoke.js
```diff
@@ -33,6 +33,33 @@ describe('React Intl mixin', function () {
...
+                number: {
+                    currency: {
+                        style: 'currency',
+                        minimumFractionDigits: 2
+                    }
+                }
+            },
+
+            format: 'currency',
+            currency: price.currency
```
ericf: This example shows how the props are being merged with the named format, because if they weren't `Intl.NumberFormat` would throw because of the missing `currency` value.

caridy:
This is perfect, easy that I initially thought! +1

PR #64, opened 2015-01-16
Rename components to Intl*

### LINKED ISSUE #60: Number/Date/Mixin components exposed as IntlNumber/IntlDate/IntlMixin
This is a feature request to export the Number/Date/Mixin components as IntlNumber/IntlDate/Mixin.

I love using the ES6 object destructuring syntax, but when I'm consuming react-intl components I cannot use it because it would "hide" the built in Number/Date types. Mixin is slightly different in that there is no conflict -- I just always rename it to IntlMixin to make it a bit more clear about it's use.

I'd love to be able to do the following.

```javascript
var { IntlNumber, IntlDate, IntlMixin } = require('react-intl');
```

Instead of

```javascript
var ReactIntl = require('react-intl'),
  IntlNumber = ReactIntl.Number,
  IntlDate = ReactIntl.Date,
  IntlMixin = ReactIntl.Mixin;
```

And if you end up going this route, I'd probably prefix the other ones with Intl for consistency.

ericf:
Fixes #60

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1,31 +1,30 @@
 /* jshint esnext: true */
 
 import {
-    Mixin,
-    Date,
-    Time,
-    Relative,
-    Number,
-    Message,
-    HTMLMessage,
+    IntlMixin,
+    IntlDate,
+    IntlTime,
+    IntlRelative,
+    IntlNumber,
+    IntlMessage,
+    IntlHTMLMessage,
     __addLocaleData
 } from './react-intl';
 
 export default {
-    Mixin      : Mixin,
-    Date       : Date,
-    Time       : Time,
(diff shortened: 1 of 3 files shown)


yahoocla:
CLA is valid!

caridy:
what about Format?, if we do Format instead of Intl, it will even facilitate the transition to FormatJS library in the future, isn't it?

ericf:
Yeah, honestly I'm on the fence about this. Since this is a component my initial thought is to use a noun-y name, whereas if it's a method then it would be a verb-y name. I will write up some more details on this decision and some other things I've been thinking as sort of an RFC.

ericf:
Closing in favor of #67

Issue #65, opened 2015-01-16
[RFC] React Intl Component Details


ericf:
A core principal behind [FormatJS][] is providing a declarative way to format data and strings by providing high-level integrations with template/component libraries. The first version of React Intl only have an imperative API provided by `ReactIntlMixin`, but `v1.1.0` will provide a declarative way via React Intl Components. See #23 

We feel that formatting data and strings using the declarative form is superior to using the imperative JavaScript APIs, and the plan is to update all the docs to encourage people to use the React Intl Components going forward. Knowning that this is a much different way to use React Intl (luckily everything is backwards compatible) we decided to cut [RC releases][v1.1.0-rc-1] before making `v1.1.0` final.

So far the feedback from people who've been testing out the `v1.1.0-rc-*` releases has been great! After reflecting on this feedback we still have some open questions that we'd like to settle before cutting the final `v1.1.0` release…

### 1. How should the React Intl Components be named?

@pselden raised a concern about the name of React Intl's module exports for each of the components in #60. Currently on `master` the exports are named assuming that they will be accessed on the `ReactIntl` namespace object; i.e., `ReactIntl.Number`.

While the current naming scheme makes sense when it's written as above, it quickly breaks down in a few ways…

#### Components with '.'s Feel Awkward to Use

Since JSX compiles down into JavaScript function calls it's perfectly valid to use '.'s when referencing components; e.g., `<ReactIntl.Number>`, but there's a worry that this might feel odd to use since JSX looks like HTML and you don't see '.'s used in HTML tag names.

#### ES6 Destructuring and Module Imports

@pselden mentions in #60 how he's been using ES6 destructuring assignments as loving it. But there's a problem with this and the current naming scheme for React Intl Components:

```js
var {Number, Date, Mixin} = require('react-intl');
```

**Built-ins are shadowed!** This is horrible, especially for `Date` which is commonly accessed in JavaScript application code. ES6 Module imports pose a similar problem, but it can be remedied using `as`:

```js
import {
    Number as IntlNumber,
    Date as IntlDate,
    Mixin as IntlMixin
} from 'react-intl';
```

While this solves the shadowing problem, it adds a ton of extra typing that will make React Intl annoying to use via ES6 Modules.

For whatever naming scheme that's chosen, we'd like to put more weight behind one with ergonomics that fit with the future direction of JavaScript; i.e. we like the ES6 shorthand destructuring syntax.

#### [Proposal] Prefixed Exports with `Intl`

[...]

PR #66, opened 2015-01-20
Change React Intl Components to be self-closing

### LINKED ISSUE #26: Strategies for using JSX in messages?
I've run into a few cases where my messages need inline JSX (like a link or presentational markup). I'm handling that now by chopping messages up around the JSX, but it's a bummer to lose the overall translation context. I'm curious if you have employed any other approaches to handle this situation?

ericf:
This modifies all of the React Intl Components to be self-closing instead of relying on the nested child to be the value of interest. All of the components (beside Message and HTMLMessage) have a `value` prop:

```jsx
<FromattedNumber value={1000} />
```

The Message and HTMLMessage components use a `message` prop instead of `value`. Both of these components now support React Elements as prop values! This makes formatting message _much_ more powerful, and for the Message component, React will be able to use virtual DOM diffing. For the HTMLMessage component there will be a perf hit since the entire message must end up as a big HTML string that's then injected into the DOM via `dangerouslySetInnerHTML`.

```jsx
var msg = "Posted {ago}, {likes, plural, one{# like} other{# likes}}";

<FormattedMessage message={msg}
   ago={<FormattedRelative value={post.date} />}
   likes={post.likes} />
```

Details of the discussion on using nested children vs. self-closing elements, and supporting React Elements in messages is in #65.

Fixes #26

### DIFF
--- a/src/components/message.js
+++ b/src/components/message.js
@@ -9,11 +9,70 @@ var IntlMessage = React.createClass({
     displayName: 'IntlMessage',
     mixins     : [IntlMixin],
 
+    propTypes: {
+        tagName: React.PropTypes.string,
+        message: React.PropTypes.string.isRequired
+    },
+
+    getDefaultProps: function () {
+        return {tagName: 'span'};
+    },
+
     render: function () {
         var props   = this.props;
-        var message = props.children;
+        var tagName = props.tagName;
+        var message = props.message;
 
-        return React.DOM.span(null, this.formatMessage(message, props));
+        // Creates a token with a random guid that should not be guessible or
+        // conflict with other parts of the `message` string.
+        var guid     = Math.floor(Math.random() * 0x10000000000).toString(16);
+        var token    = '@__ELEMENT_' + guid + '__@';
+        var elements = [];
+
+        // Iterates over the `props` to keep track of any React Element values
+        // so they can be represented by the `token` as a placeholder when the
+        // `message` is formatted. This allows the formatted message to then be
+        // broken-up into parts with references to the React Elements inserted
+        // back in.
+        var values = Object.keys(props).reduce(function (values, name) {
+            var value = props[name];
+
+            if (React.isValidElement(value)) {
+                values[name] = token;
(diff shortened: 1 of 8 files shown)


yahoocla:
CLA is valid!

src/components/message.js
```diff
@@ -9,11 +9,70 @@ var IntlMessage = React.createClass({
...
+        // so they can be represented by the `token` as a placeholder when the
+        // `message` is formatted. This allows the formatted message to then be
+        // broken-up into parts with references to the React Elements inserted
+        // back in.
+        var values = Object.keys(props).reduce(function (values, name) {
+            var value = props[name];
+
+            if (React.isValidElement(value)) {
+                values[name] = token;
+                elements.push(value);
```
ericf: Realized there's a bug here since there's no reference to _which_ elements should be inserted back _where_.

src/components/message.js
```diff
@@ -9,11 +9,67 @@ var IntlMessage = React.createClass({
...
+
+        // Formats the `message` with the `values`, including the `token`
+        // placeholders for React Element values.
+        var formattedMessage = this.formatMessage(message, values);
+
+        // Split the message into parts, and replace the token placeholder parts
+        // with references to the React Element values captured above. This
+        // approach allows messages to render with React Elements while keeping
+        // React's virtual diffing working properly.
+        var children = formattedMessage.split(tokenRegex).map(function (part) {
```
caridy: hmm, I didn't know this was possible, pretty cool:

```
var tokenRegex = new RegExp('(@__ELEMENT-' + guid + '-\\d__@)', 'g');
"foo@__ELEMENT-1-1__@bar".split(tokenRegex)
[ 'foo',
  '@__ELEMENT-1-1__@',
  'bar' ]
```

caridy:
+1

jbelcher:
@ericf +1. We actually ended up creating our own `Message` component that wraps `IntlMessage` just so we can use this type of interface. It'll be great to have built-in support for that.

PR #67, opened 2015-01-21
Prefix Components with "Formatted"

### LINKED ISSUE #60: Number/Date/Mixin components exposed as IntlNumber/IntlDate/IntlMixin

ericf:
This adds "Formatted" as a prefix to all React Intl Components.
See: #65 for discussion

Closes #64
Fixes #60

### DIFF
--- a/example/src/components.js
+++ b/example/src/components.js
@@ -2,29 +2,30 @@
 
 (function (global) {
 
-var IntlDate        = ReactIntl.Date;
-var IntlTime        = ReactIntl.Time;
-var IntlRelative    = ReactIntl.Relative;
-var IntlNumber      = ReactIntl.Number;
-var IntlMessage     = ReactIntl.Message;
-var IntlHTMLMessage = ReactIntl.HTMLMessage;
+var IntlMixin            = ReactIntl.IntlMixin;
+var FormattedDate        = ReactIntl.FormattedDate;
+var FormattedTime        = ReactIntl.FormattedTime;
+var FormattedRelative    = ReactIntl.FormattedRelative;
+var FormattedNumber      = ReactIntl.FormattedNumber;
+var FormattedMessage     = ReactIntl.FormattedMessage;
+var FormattedHTMLMessage = ReactIntl.FormattedHTMLMessage;
 
 var Another = React.createClass({
-  mixins: [ReactIntlMixin],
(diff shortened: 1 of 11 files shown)


yahoocla:
CLA is valid!

caridy:
My only issue is with `ReactIntl.IntlMixin`, which seems awkward, where Mixin is not that big of a problem, like it is Number or Date

caridy:
other than that, it LGTM

ericf:
@caridy my thoughts around `IntlMixin` vs. `Mixin` as the export name are that when looking forward it will be common for people to use destructuring and module imports in ES6, and I think that changes how me might want to name our exports.

```js
var {IntlMixin, FormattedNumber} = require('react-intl);
```

```js
import {IntlMixin, FormattedNumber} from 'react-intl';
```

In these two cases it seems preferable to have `IntlMixin` as the name of the prop/export. Otherwise when using ES6 modules you'll want to rename the named export for clarity:

```js
import {Mixin} from 'react-intl';

export default class extends Component {
    mixins: [Mixin], // What is mixin is "Mixin" again?
    render: function () {
        // ...
    }
}
```

Since it won't be clear what the `Mixin` is when looking at the code without referring to the `import` statement, people will probably end up doing:

```js
import {Mixin as IntlMixin} from 'react-intl';
```

So I thought it makes sense to save the need for using `as` and just name the export `IntlMixin` from the start. What do you think? Do you agree when thinking of it this way, or am I being silly :) ?

caridy:
ok, let's roll with IntlMixin

PR #69, opened 2015-01-22
Fix/Imporove React Element tokenization in messages

ericf:
This fixes an issue with the message tokenization process in `<FormattedMessage>` by supporting more than 9 React Elements passed as prop values.

The rendered message is also improved by filtering out any empty string parts so empty `<span>`s are not created in the DOM.

### DIFF
--- a/src/components/message.js
+++ b/src/components/message.js
@@ -23,10 +23,10 @@ var FormattedMessage = React.createClass({
         var tagName = props.tagName;
         var message = props.message;
 
-        // Creates a token with a random guid that should not be guessible or
+        // Creates a token with a random guid that should not be guessable or
         // conflict with other parts of the `message` string.
         var guid       = Math.floor(Math.random() * 0x10000000000).toString(16);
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

src/components/message.js
```diff
@@ -23,10 +23,10 @@ var FormattedMessage = React.createClass({
         var tagName = props.tagName;
         var message = props.message;
 
-        // Creates a token with a random guid that should not be guessible or
+        // Creates a token with a random guid that should not be guessable or
         // conflict with other parts of the `message` string.
         var guid       = Math.floor(Math.random() * 0x10000000000).toString(16);
-        var tokenRegex = new RegExp('(@__ELEMENT-' + guid + '-\\d__@)', 'g');
+        var tokenRegex = new RegExp('(@__ELEMENT-' + guid + '-\\d+__@)', 'g');
```
caridy: ha, my bad, I should have spotted that.

caridy:
+1

src/components/message.js
ericf: No worries, the main thing was it was spotted before the release :)

PR #70, opened 2015-01-23
Reduce content in README in favor of the site

ericf:
We want one source of truth for the docs, the website, so this reduces the README down and links off to the FormatJS website.


yahoocla:
CLA is valid!

caridy:
+1

tripp:
:+1:

PR #88, opened 2015-03-09
Update react dependency to: ">=0.11.2 <1.0.0"

### LINKED ISSUE #79: Move react to peerDependencies
I'm using react with browserify and had some problems today with this nasty bug

> Uncaught TypeError: Cannot read property 'firstChild' of undefined
/~/react-router/~/react/lib/ReactMount.js?:606

Luckily I found the solution at [StackOverflow](http://stackoverflow.com/questions/27153166/typeerror-when-using-react-cannot-read-property-firstchild-of-undefined).

> For example, if you npm install a package that requires a different React version and puts it into dependencies instead of peerDependencies, it might install a separate React into node_modules/<some library using React>/node_modules/react.
Two different Reacts won't play nicely together (at least yet).

I removed `./node_modules/react-intl/node_modules/react` and now everything works fine again.

Please update your [package.json](https://github.com/yahoo/react-intl/blob/master/package.json) and move `react` to `peerDependencies`.

ericf:
This brings in support for react@0.13, plus any future pre-1.0 versions. React has a great back-compat track record, so it should be safe to do this. If an issue arrises, we can always back this down in a patch release.

Closes #79

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "intl-format-cache": "2.0.4",
     "intl-messageformat": "1.0.4",
     "intl-relativeformat": "1.0.3",
-    "react": ">=0.11.2 <0.13.0"
+    "react": ">=0.11.2 <1.0.0"
   },
   "devDependencies": {
     "es5-shim": "^4.0.3",


yahoocla:
CLA is valid!

caridy:
:+1:

PR #90, opened 2015-03-10
Change `propsTypes` to `propTypes` in mixin.js

lencioni:
This looks like a typo to me. It seems that it was intended to be `propTypes`.


yahoocla:
CLA is valid!

ericf:
Good catch, thanks!

PR #127, opened 2015-06-18
Exclude tests file from npm package

roderickhsiao:
@ericf 

I am not sure should we exclude /src folder and just keep /dist as well.

Thanks

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
 example/
 tmp/
+tests/
 Gruntfile.js


yahoocla:
CLA is valid!

caridy:
+1 on ignoring tests

but we can't ignore src because you might need to use it in the ES6 land in the future, check this: https://github.com/yahoo/react-intl/blob/master/package.json#L24

roderickhsiao:
sure, thanks, then I will just keep the PR as is :)

PR #128, opened 2015-06-23
Update package.son to be a peerDependency

crafterm:
Hi There,

This source for this fix was originally the react dev tools in Chrome failing to recognise components on the page, and showing an empty top-level tag.

The cause appeared to be 2 versions of react being present in my webpack created bundle. One the app specified version, and another being a beta 0.14 version that I tracked down to be a dependency of react-intl.

Have updated the package.json to specify react as a peer dependency so that bundling tools such as webpack, etc, don’t bundle react-intl’s version of react as an additional copy.

After this change, and refreshing the react-intl node-modules directory, this fixed the original issue.

Cheers,

Marcus

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,8 +29,10 @@
   "dependencies": {
     "intl-format-cache": "2.0.4",
     "intl-messageformat": "1.1.0",
-    "intl-relativeformat": "1.1.0",
-    "react": ">=0.11.2 <1.0.0"
+    "intl-relativeformat": "1.1.0"
+  },
+  "peerDependencies": {
+    "react": "^0.13.3"
   },
   "devDependencies": {
     "es5-shim": "^4.1.0",


caridy:
We have discussed this extensibly. We are not going to change this one, you should be able to control the version of react in your package.json at the app level.

spicyj:
@caridy My understanding is that this is exactly what peer deps are for – for stateful dependencies where you want multiple dependent modules to share one dependency (in this case, React). We (React team) recommend peer dependencies for any library that depends on React. Standard is to mark `^0.13.0` or similar – you can't claim to support `0.14.x` because we might make arbitrary breaking changes in that version. If it happens to still work with 0.14, people can still use it – peer dependency versions merely warn if there's a mismatch. (We're also working to reduce the problems you currently encounter when pulling in multiple versions of React.)

caridy:
@spicyj let me try to clarify my position. I had another long discussion with @ericf yesterday about this particular issue, and there are few things we need to keep in mind:

* if you use `npm install --production` dupes are taken care of properly, but since people are bundling files using dev dependencies, this gets very tricky.
* `peerDependecies` are problematic in principle, and it seems the future of it is just to define "intent" rather than "explicit" definition. For what I can tell from the latest discussions, you will have to have those dependencies in dep or devDep in order to guarantee the pkg to be installed. See more here: https://github.com/npm/npm/issues/6565#issuecomment-74971689
* `npm 3.x` will mitigate the duping issue by flattening the structure and use dupes only when conflicts appear. (and that will be applied to devDeps as well, which is the root of all evil when using browserify/webpack).

My position is: lets wait until we have a clear picture. Ideally npm v3 will fix that for us.

spicyj:
peerDependencies won't be installed automatically, but I'm having trouble imagining when you would ever depend on react-intl but not react. Even if your app uses react 0.14.x, you would never want react-intl to pull in 0.13.x just because that's what react-intl has been tested with. You'll get two versions of React and havoc will ensue. The way React currently works, you never want dupes. (And we're working on that. Hang with us.)

Can you help me understand what situations your current `dependencies` spec is preferable?

Enet4:
Allow me to back this issue. React as a peer dependency is currently the most adequate way to have React-dependent libraries under the current behavior of `peerDependencies` in npm. Waiting for v3
means this package will remain problematic until that is released, which would unnecessarily cripple other people's work depending on it.

Like @spicyj, I would also like to understand why you are currently defining React and other bundled modules as `dependencies`, as I can think of nothing but the obvious downside: if it's bundled in development time, it's not needed it in production at all. Why is this approach "the root of all evil" then?

caridy:
@Enet4:

> Waiting for v3
means this package will remain problematic until that is released, which would unnecessarily cripple other people's work depending on it.

npm moves faster that most other pieces in the node ecosystem. Last week, the first beta of npm3 was released, and this is coming fast toward us. My preliminar tests suggest that there is no issue at all with npm3, I will keep digging more. More details here: http://blog.npmjs.org/post/122450408965/npm-weekly-20-npm-3-is-here-ish

>  if it's bundled in development time, it's not needed it in production at all.

that's not accurate, for isomorphic apps, you still need it on the server side.

> Why is this approach "the root of all evil" then?

we have many pkgs, not only the react distro, I don't want to change them all now, and having to change them again in the near future, considering that the proposed solution in this PR is not going to solve the issue in the future. Ultimately, we will have react as a peerDep and Dep as well to match npm3 behavior.

@spicyj:

> I'm having trouble imagining when you would ever depend on react-intl but not react.

you're right, under normal conditions they will have both of them defined in pkg.json, side by side.

> Even if your app uses react 0.14.x, you would never want react-intl to pull in 0.13.x just because that's what react-intl has been tested with.

We are not pulling any specific version, in fact the requirement supports a wide range of versions today: `"react": ">=0.11.2 <1.0.0"`.

> Can you help me understand what situations your current dependencies spec is preferable?

It is mostly the uncertainty of the peerDeps, internally we have many issues with peerDepedencies (keep in mind that we have our own registry and our own npm distro at yahoo). Things are getting clear up now with npm3, I personally want to hold on this until we have a clear picture.

jamesseanwright:
I agree with @spicyj, why would one install this package without React? The fact that react-intl specifies React as an explicit dependency is causing [this issue](https://github.com/facebook/react/issues/3396) in our codebase. Ideally your library should be using the parent project's version of React. What exactly are the issues that you've encountered with npm's `peerDependencies`? For now I'll have to point to my own fork of this library.

caridy:
it is coming with v2
https://github.com/yahoo/react-intl/blob/v2-dev/package.json#L36-L38

jamesseanwright:
That's excellent news, thanks @caridy!

ericf:
See: #162

Issue #130, opened 2015-06-24
support for optional tag name instead of hardcoded span element


rsamec:
I would like to suggested replacing

```js
return React.DOM.span(null, this.formatDate(value, options));
```
with support for optional tag 

```js
var tagName  = props.tagName || 'span';
return React.DOM[tagName](null, this.formatDate(value, options));
```

react-Intl components are usable than i svg <text>, that accepts only __tspan__ element.

caridy:
We have discussed this in the past, but not in the context of svg, which brings something new to the table. I will discuss this again with @ericf soon.

ericf:
Yeah I see the issue for SVG, so this makes sense to add.

caridy:
awesome.

PR #131, opened 2015-06-24
support for optional tagName for components -> FormattedDate, Formatt…

rsamec:
support for optional tagName for components instead of hardcoded __span__ element.

It solves the [issue](https://github.com/yahoo/react-intl/issues/130)

### DIFF
--- a/src/components/date.js
+++ b/src/components/date.js
@@ -24,12 +24,13 @@ var FormattedDate = React.createClass({
 
     render: function () {
         var props    = this.props;
+        var tagName  = props.tagName || 'span';
         var value    = props.value;
         var format   = props.format;
         var defaults = format && this.getNamedFormat('date', format);
         var options  = FormattedDate.filterFormatOptions(props, defaults);
 
-        return React.DOM.span(null, this.formatDate(value, options));
+        return React.DOM[tagName](null, this.formatDate(value, options));
     }
 });
 
--- a/src/components/date.js
+++ b/src/components/date.js
@@ -22,9 +22,13 @@ var FormattedDate = React.createClass({
         value : React.PropTypes.any.isRequired
     },
 
+    getDefaultProps: function () {
+        return {tagName: 'span'};
+    },
+
     render: function () {
         var props    = this.props;
-        var tagName  = props.tagName || 'span';
+        var tagName  = props.tagName;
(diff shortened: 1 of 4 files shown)


src/components/date.js
```diff
@@ -24,12 +24,13 @@ var FormattedDate = React.createClass({
 
     render: function () {
         var props    = this.props;
+        var tagName  = props.tagName || 'span';
```
endragor: Specify 'span' as a [default value](https://facebook.github.io/react/docs/reusable-components.html#default-prop-values).

ericf:
Just leaving a note here that `tagName` was added to really only help edge cases like SVG, and if additional styling is desired, a developer should wrap the React Intl components.

gpbl:
@ericf maybe a warning when using `tagName` with values other than `tspan` or `span`?

elCarda:
@ericf Please release it as a minor version? Or any release in roadmap soon?

ericf:
Trying to determine if `component` is better than `tagName` and will help us support React Native as well…

Also, in the [`v2-dev` branch](https://github.com/yahoo/react-intl/tree/v2-dev) all of the components now support the function-as-child pattern which helps to mitigate this.

SimenB:
Chiming in months later with a request to change it to `component` as noted above. That more closely mirrors React's own

ericf:
@SimenB What's your use case? I won't be adding support for passing various props to the underlying component that's rendered; if you need to customize rendering in that way you can pass a function as the child.

SimenB:
No use case (we already wrap the `intl`-components), this suggestion was just to rename your prop `tagName` to `component` like you suggested above, for the sake of consistency with React itself

ericf:
What do you use `tagName`/`component` for? To you ever assign a value to it?

SimenB:
We don't use it (or, we did, until I watched "Dom as second-class citizen"). My comment was just on the fact  that you named your prop something other than React did, which makes it a bit less intuitive (even though I know its use is discouraged).

SimenB:
Of course, transition group is a container and requires children which intl does not, so it's not the exact same thing

ericf:
The driving force for me to support this is rendering in React Native and SVGs which don't have `<span>`.

PR #134, opened 2015-06-30
Added `className` and `style` props to all components

### LINKED ISSUE #72: Pass through style and className props to components
i often find myself needing to style (via style or className props) react-intl components
i see none of your components use the style or className props so far so i could submit a pull request to support that if you're fine with it.

endragor:
Closes #72.

### DIFF
--- a/src/components/date.js
+++ b/src/components/date.js
@@ -18,18 +18,25 @@ var FormattedDate = React.createClass({
     },
 
     propTypes: {
-        format: React.PropTypes.string,
-        value : React.PropTypes.any.isRequired
+        format   : React.PropTypes.string,
+        value    : React.PropTypes.any.isRequired,
+        className: React.PropTypes.string,
+        style    : React.PropTypes.object
     },
 
     render: function () {
-        var props    = this.props;
-        var value    = props.value;
-        var format   = props.format;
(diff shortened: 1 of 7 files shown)


yahoocla:
CLA is valid!

caridy:
dupe of https://github.com/yahoo/react-intl/pull/131, tagName is being discussed there. classname and style are out of the question for me because you can achieve the same by wrapping the component, without blacklisting potential tokens to be used in ICU messages.

endragor:
Then could you please leave this comment in #72 and close it, too? This is not really a dupe of #131.
Yes, I may achieve the same by wrapping, but the layout (DOM) becomes way more verbose and ugly, if I wrap the FormattedMessage everywhere.

endragor:
I'll then just add wrapper component in my own project, but it's still worth commenting and closing #72, so that your opinion is clear and noone else tries to do this work in vain.

caridy:
yeah, apologies for not being clear on that issue, I will sync up with @ericf about this.

ericf:
If there was a way to output a pure DOM text node, that's what we'd do from these components; but since there isn't, we have to use a `<span>` which is the more inert element.

We should add docs to the website which explain how to style these components via wrapping them.

PR #135, opened 2015-06-30
Update Readme

dropfen:
Add a Note, that you can also provide a specific rule for a specific value


yahoocla:
CLA is valid!

ericf:
Thanks!

PR #144, opened 2015-08-12
set lang to components

sergiodxa:


ericf:
This is not something we plan to add. People are already too confused with `locales` being plural. I don't think we should add another similar property.

In the [`v2-dev` branch](https://github.com/yahoo/react-intl/tree/v2-dev) I've added a new `<IntlProvider>` component with [these props](https://github.com/yahoo/react-intl/blob/b78adf52991db7bbb283fde00778fb06245c184a/src/types.js#L5-L12). I've decided to make `locale` singular, and `defaultLocale` is used when formatting needs to fallback, e.g. when translations fail to load.

PR #153, opened 2015-08-26
Explicitly add uglify-js

ryan1234:
I don't have uglify-js installed globally, so had to add this to get `npm install` to work for the first time.


ericf:
Ah cool, my mistake for not adding it. Thanks!

PR #155, opened 2015-08-28
[v2] Refactor imperative API

ericf:
This reduces React Intl's API surface and simplifies the imperative API.

There is now a `defineMessage()` export which serves as the hook for the static analysis message extraction via babel-plugin-react-intl.

The `this.context.intl` object now contains the format functions.

These changes make the imperative API less complex as well as more flexible for defining messages and using them more than once.

/cc @caridy @redonkulus @mridgway


caridy:
:+1:

PR #161, opened 2015-09-11
[v2] Fix `defaultMessage` in example

mattikl:
Minor detail, but can be confusing to someone learning the message syntax from the example.


mattikl:
The Travis build is failing due to a bug in eslint 1.4.0 and you also need to change react version to 0.14.0-rc1 in `package.json`. So many things going on at this very moment :)

ericf:
Good catch! I cherry-picked this into `master`. Thanks!

PR #168, opened 2015-09-17
Add injectIntl() HOC

ericf:
This will inject the `intl` context object created by the `<IntlProvider>` as a prop on the wrapped component. Using the HOC factory function allivates the need for `context` to be a part of the
public API.


juandopazo:
`provideIntl`?

ericf:
I think that's too close to `<IntlProvider>`. I went with "inject" because it's more like Dependency Injection.

mridgway:
lgtm

PR #169, opened 2015-09-17
Update dist/locale-data/ to be UMD files

ericf:
This keeps with the theme of v2 where locale data is provided as modules, instead of causing side-effects to the ReactIntl library.

With the change, the locale data files in `dist/` will expose the data at: `ReactIntlLocaleData.<lang>`. This decouples the loading of the locale data files from the loading of the library.


mridgway:
lgtm

PR #171, opened 2015-09-18
Replace defineMessage() with defineMessages()

ericf:
This is a new function that allow multiple messages to be defined (and extracted by the Babel plugin.) The bigger question is whether the singular form `defineMessage()` should be removed in favor of just having this one.

### DIFF
--- a/src/react-intl.js
+++ b/src/react-intl.js
@@ -20,12 +20,6 @@ export {default as FormattedHTMLMessage} from './components/html-message';
 
 export {intlShape} from './types';
 
-// TODO: Can we get rid of this nad only go with the multiple version?
-export function defineMessage(messageDescriptor) {
-    // TODO: Type check in dev? Return something different?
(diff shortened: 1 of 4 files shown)


jarnoux:
:+1: :relaxed:

mridgway:
:+1: 

I would be fine with removing `defineMessage`.

src/react-intl.js
```diff
@@ -20,12 +20,17 @@ export {default as FormattedHTMLMessage} from './components/html-message';
 
 export {intlShape} from './types';
 
-// TODO Should there be a `defineMessages()` function that takes a collection?
+// TODO: Can we get rid of this nad only go with the multiple version?
```
mridgway: Type: `nad` -> `and`

ericf:
Okay, I've removed `defineMessage()`, singular to avoid confusion.

Issue #177, opened 2015-10-01
v2 does not install with Node 4.1 / Npm3


baer:
I just tried running this and I found two issues:

1. NPM 3.x does NOT install peerDependencies for you. As such you will need to add `jsdom` to your package.json to avoid errors related to the following error from npm:
```
UNMET PEER DEPENDENCY jsdom@*
```

2. It looks like one of the dependencies (`contextify`) does not build under Node 4.1.x. I'm about to try with the 0.10.x line but this is an issue worth solving regardless.

```
npm install
npm WARN prefer global nave@0.5.2 should be installed with -g

> contextify@0.1.14 install /Users/ebaer/dev/open-source/react-intl/node_modules/contextify
> node-gyp rebuild

  CXX(target) Release/obj.target/contextify/src/contextify.o
In file included from ../src/contextify.cc:3:
../node_modules/nan/nan.h:261:25: error: redefinition of '_NanEnsureLocal'
NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Local<T> val) {
                        ^
../node_modules/nan/nan.h:256:25: note: previous definition is here
NAN_INLINE v8::Local<T> _NanEnsureLocal(v8::Handle<T> val) {
                        ^
../node_modules/nan/nan.h:661:13: error: no member named 'smalloc' in namespace 'node'
    , node::smalloc::FreeCallback callback
      ~~~~~~^
../node_modules/nan/nan.h:672:12: error: no matching function for call to 'New'
    return node::Buffer::New(v8::Isolate::GetCurrent(), data, size);
           ^~~~~~~~~~~~~~~~~
...
```

```
npm ERR! Darwin 14.5.0
npm ERR! argv "/Users/ebaer/.nvm/versions/node/v4.1.1/bin/node" "/Users/ebaer/.nvm/versions/node/v4.1.1/bin/npm" "install"
npm ERR! node v4.1.1
npm ERR! npm  v3.3.4
npm ERR! code ELIFECYCLE

npm ERR! contextify@0.1.14 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the contextify@0.1.14 install script 'node-gyp rebuild'.
npm ERR! This is most likely a problem with the contextify package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild
npm ERR! You can get their info via:
npm ERR!     npm owner ls contextify
[...]

PR #178, opened 2015-10-01
Update jsDOM to the latest version

baer:
The jsDOM being included was 3 major versions back from the lates and was causing build failures in node 4.1 and 0.10. Updating the version corrects the issue.

Closes https://github.com/yahoo/react-intl/issues/177

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "expect": "^1.9.0",
     "formatjs-extract-cldr-data": "^1.1.1",
     "intl": "^1.0.0",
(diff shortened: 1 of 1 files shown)


baer:
Your CLA server is down :/

ericf:
I'm not using this dep yet, so I think instead of upgrading we can just remove it for the time being.

baer:
Yet? What would you use jsDOM for in an intl lib?

ericf:
If it's used it will only be for testing.

PR #181, opened 2015-10-08
Downgrade peer dependency to React 0.13

jnak:
The V2 RFC mentions that React 0.13 should be supported


ericf:
Thanks for proposing this as fixing a mistake, but it's actually intentional.

I'll update the RFC, only 0.14+ will be supported. I tried supporting both, but when integrating into apps I ran into `peerDependency` hell.

PR #191, opened 2015-10-21
Stabilize "now" time and "tick" relative times

### LINKED ISSUE #186: <FormattedRelative /> React attempted to reuse markup in a container but the checksum was invalid
I'm trying to use `<FormattedRelative>` in a universal javascript environment and I'm not sure how to make this error go away:

```
Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) 0.$0.0.1.2.1.1.1.0">8 seconds ago</span>
 (server) 0.$0.0.1.2.1.1.1.0">5 seconds ago</span>
```

ericf:
This adds a new `initialNow` prop to `<IntlProvider>` to stabilize the reference time when formatting relative times during the initial render. This prop should be set when rendering a universal React app
on the server and client so the initial client render will match the server's checksum.

On the server, `Date.now()` should be captured before calling `ReactDOM.renderToString()` and provided to `<IntlProvider>`. This "now" value needs to be serialized to the client so it can also pass the same value to `<IntlProvider>` when it calls `React.render()`.

Relatives times formatted via `<FormattedRelative>` will now "tick" and stay up to date over time. The `<FormattedRelative>` component's `now` prop has been renamed, `initialNow` to match and override the same prop on `<IntlProvider>`. It also has a new `updateInterval` prop which accepts a number of milliseconds for the maximum speed at which relative times should be updated (defaults to 10 seconds).

Special care has been taken in the scheduling algorithm to display accurate information while reducing unnecessary re-renders. The algorithm will update the relative time at its next "interesting" moment; e.g., "1 minute ago" to "2 minutes ago" will use a delay of 60 seconds even if `updateInterval` is set to 1 second.

Fixes #186

### DIFF
--- a/src/components/relative.js
+++ b/src/components/relative.js
@@ -8,21 +8,112 @@ import React, {Component, PropTypes} from 'react';
 import {intlShape, relativeFormatPropTypes} from '../types';
 import {invariantIntlContext, shouldIntlComponentUpdate} from '../utils';
 
+const SECOND = 1000;
+const MINUTE = 1000 * 60;
+const HOUR   = 1000 * 60 * 60;
+const DAY    = 1000 * 60 * 60 * 24;
+
+// The maximum timer delay value is a 32-bit signed integer.
+// See: https://mdn.io/setTimeout
+const MAX_TIMER_DELAY = 2147483647;
+
+function selectUnits(delta) {
+    let absDelta = Math.abs(delta);
+
+    if (absDelta < MINUTE) {
+        return 'second';
+    }
+
+    if (absDelta < HOUR) {
+        return 'minute';
+    }
+
+    if (absDelta < DAY) {
+        return 'hour';
+    }
+
+    // The maximum scheduled delay will be measured in days since the maximum
+    // timer delay is less than the number of milliseconds in 25 days.
+    return 'day';
+}
+
+function getUnitDelay(units) {
+    switch (units) {
+    case 'second': return SECOND;
+    case 'minute': return MINUTE;
+    case 'hour'  : return HOUR;
+    case 'day'   : return DAY;
+    default      : return MAX_TIMER_DELAY;
+    }
+}
+
 export default class FormattedRelative extends Component {
     constructor(props, context) {
         super(props, context);
         invariantIntlContext(context);
+
+        let now = isFinite(props.initialNow) ?
+                Number(props.initialNow) : context.intl.now();
+
+        // `now` is stored as state so that `render()` remains a function of
+        // props + state, instead of accessing `Date.now()` inside `render()`.
+        this.state = {now};
+    }
+
+    scheduleNextUpdate(props, state) {
+        const {updateInterval} = props;
+
+        // If the `updateInterval` is falsy, including `0`, then auto updates
+        // have been turned off, so we bail and skip scheduling an update.
+        if (!updateInterval) {
+            return;
+        }
+
+        let delta = Number(props.value) - state.now;
+        let units = props.units || selectUnits(delta);
+
+        let unitDelay     = getUnitDelay(units);
+        let unitRemainder = Math.abs(delta % unitDelay);
+
+        // We want the largest possible timer delay which will still display
+        // accurate information while reducing unnecessary re-renders. The delay
+        // should be until the next "interesting" moment, like a tick from
+        // "1 minute ago" to "2 minutes ago" when the delta is 120,000ms.
+        let delay = delta < 0 ?
+            Math.max(updateInterval, unitDelay - unitRemainder) :
+            Math.max(updateInterval, unitRemainder);
+
+        clearTimeout(this._timer);
+
+        this._timer = setTimeout(() => {
+            this.setState({now: this.context.intl.now()});
+        }, delay);
     }
 
     shouldComponentUpdate(...next) {
         return shouldIntlComponentUpdate(this, ...next);
     }
 
+    componentWillUpdate(nextProps, nextState) {
+        this.scheduleNextUpdate(nextProps, nextState);
+    }
+
+    componentDidMount() {
+        this.scheduleNextUpdate(this.props, this.state);
+    }
+
+    componentWillUnmount() {
+        clearTimeout(this._timer);
+    }
+
     render() {
         const {formatRelative} = this.context.intl;
-        const props            = this.props;
+        const {props, state}   = this;
 
-        let formattedRelative = formatRelative(props.value, props);
+        let formattedRelative = formatRelative(props.value, {
+            ...props,
+            ...state,
+        });
 
         if (typeof props.children === 'function') {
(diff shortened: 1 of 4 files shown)


ericf:
I'm open to bikeshedding `initialNow` and `updateInterval` names.

mridgway:
I'm fine with those names and the default. Looks good to me. :+1:

src/components/relative.js
```diff
@@ -8,21 +8,112 @@ import React, {Component, PropTypes} from 'react';
...
+const MINUTE = 1000 * 60;
+const HOUR   = 1000 * 60 * 60;
+const DAY    = 1000 * 60 * 60 * 24;
+
+// The maximum timer delay value is a 32-bit signed integer.
+// See: https://mdn.io/setTimeout
+const MAX_TIMER_DELAY = 2147483647;
+
+function selectUnits(delta) {
+    let absDelta = Math.abs(delta);
```
johanneslumpe: Nitpick: this should be `const`

src/components/relative.js
```diff
@@ -8,21 +8,112 @@ import React, {Component, PropTypes} from 'react';
...
+    default      : return MAX_TIMER_DELAY;
+    }
+}
+
 export default class FormattedRelative extends Component {
     constructor(props, context) {
         super(props, context);
         invariantIntlContext(context);
+
+        let now = isFinite(props.initialNow) ?
```
johanneslumpe: `const`

src/components/relative.js
```diff
@@ -8,21 +8,112 @@ import React, {Component, PropTypes} from 'react';
...
+
+    scheduleNextUpdate(props, state) {
+        const {updateInterval} = props;
+
+        // If the `updateInterval` is falsy, including `0`, then auto updates
+        // have been turned off, so we bail and skip scheduling an update.
+        if (!updateInterval) {
+            return;
+        }
+
```
johanneslumpe: All these are not modified -> `const`

src/components/relative.js
```diff
@@ -8,21 +8,112 @@ import React, {Component, PropTypes} from 'react';
...
+        clearTimeout(this._timer);
+    }
+
     render() {
         const {formatRelative} = this.context.intl;
-        const props            = this.props;
+        const {props, state}   = this;
 
-        let formattedRelative = formatRelative(props.value, props);
+        let formattedRelative = formatRelative(props.value, {
```
johanneslumpe: And another `const`

johanneslumpe:
@ericf Haven't tested the code yet, but looks good to me apart from my `const` nitpicking ;)

caridy:
about the ticking... (`updateInterval`) I think that can be optimized based on the previous value, saying, if it renders as `1 week ago`, there is probably not need to double check that in the next 24 hours. In other words, we can simply rely on the value to be render to predict when this value should be re-rendered. I know that react will opt out of the operation if we re-render it every 10 seconds, but still, we are carrying an operation that is not really needed.

I see two options to solve this... to get the low level api to expose metadata about the formatted value (similar to what we are trying to do with datatimeformat and `fieldPositions`), or sort-of re-implement part of the logic in `FormattedRelative>` so it knows when to re-render itself.

ericf:
> Right now we would have to change the value for every single component if we wanted to tick every second.

@johanneslumpe After thinking about this, I realized the idiomatic React solution would be to wrap `<FormattedRelative>` with your own component that has your customizations, or it would naturally exist inside of some custom component's `render()`. If your UI is displaying a stream (like Twitter), you'd likely have a `<StreamItem>` component. So there might be 100 instances on the page, but `updateInterval={1000}` would only be declared once.

> Or would the solution here be to just create our own component which wraps FormattedRelative and sets the update interval?

Yeah I'm thinking this for now. If it becomes a desired feature, then it can be added to `<IntlProvider>` later.

ericf:
> I think that can be optimized based on the previous value, saying, if it renders as 1 week ago, there is probably not need to double check that in the next 24 hours. In other words, we can simply rely on the value to be render to predict when this value should be re-rendered. I know that react will opt out of the operation if we re-render it every 10 seconds, but still, we are carrying an operation that is not really needed.

@caridy This is already the algorithm I've implemented in this PR, I tried to explain it in the description at the top.

> I see two options to solve this... to get the low level api to expose metadata about the formatted value (similar to what we are trying to do with datatimeformat and fieldPositions), or sort-of re-implement part of the logic in FormattedRelative> so it knows when to re-render itself.

There's no need for this. Part of Intl RelativeFormat's API is the set of units it works with. I already have those encoded in `<FormattedRelative>`'s `units` `propType` definition. The one thing I'm ignoring is Intl RelativeFormat's thresholds, and instead compute the standard ms per `second`, `minute`, `hour`, and `day`.

`updateInterval` is essentially to set the **maximum** speed at which the relative time will be re-rendered. I added this configuration option for the following reasons:

1. To be able to turn off auto-updating relative times, `updateInterval={0}`.
2. To control how fast relative times under 1 minute "tick". For a real time app, like Twitter, you'd likely want to see: "now"…"1 second ago"…"2 seconds ago"…etc. While an app like Mail might find this distracting and only want to update every 10 seconds: "now"…"10 seconds ago"…etc.

So maybe `updateInterval` isn't the best? But I feel it gives developers some good control over the **maximum** update speed to control how "real-time" the app appears.

src/components/relative.js
ericf: This is my current coding style. I use `const` to represent constant bindings to constant values, generally which are static and not computed.

In places where I'm computing values, like this case, I've been using `let`. I may change my coding style in the future, but I'd want to apply it repo-wide.

ericf: I might switch to [this style](http://www.2ality.com/2015/02/es6-scoping.html#when_should_i_use_%3Ccode%3Elet%3C%2Fcode%3E_when_%3Ccode%3Econst%3C%2Fcode%3E%3F), but will do so in another PR.

caridy:
@ericf maybe calling it `minInterval`, which is really what it is.

ericf:
Yeah `minUpdateInterval`, but what's awkward is a value that represents the fastest something will update and the maximum frequency, but uses the term "min".

ericf:
@caridy After thinking through the naming of the prop more, I think `updateInterval` is superior. It matches the React model of being declarative, but having the framework do something smart under to the hood for performance.

I think the reason you were concern about this props name is because you were in the context of reviewing these changes and concerned about the implementation details. But for a user, they shouldn't be concerned about the implementation details; and I'll document that the scheduling algorithm schedules the updates based on the `updateInterval` and interesting moments.

caridy:
fair enough!

PR #206, opened 2015-11-04
fix(react-intl/examples) fix wrong paths in translations examples

BenNG:


BenNG:
Ok I understand...
I usually don't clone then build the project to test the examples I just copy/paste them 
My bad !

PR #208, opened 2015-11-05
Update README.md

zkirill:
Add `npm install` before `npm run build` to make sure that dependencies are installed.


ericf:
I removed the need for this in https://github.com/yahoo/react-intl/commit/c9fccf0ae5b508304a722ca731694108386e33cb

PR #217, opened 2015-11-14
Improve support for nested <IntlProvider>s

ericf:
This makes it easier and more intuitive to nest `<IntlProvider>` elements by looking for `context.intl` and merging it with its own `props`.

A common use case is for apps whose UIs are split among multiple packages, each with their own translated messages. These chunks of UI can now be wrapped with a nested `<IntlProvider>` configured to only expose the translated message for that package's root component.

**See:** `examples/nested/` for a simple example using nested `<IntlProvider>` elements.

### DIFF
--- a/examples/hello-world/package.json
+++ b/examples/hello-world/package.json
@@ -5,6 +5,7 @@
   "private": true,
   "main": "index.js",
   "scripts": {
+    "clean": "../../node_modules/.bin/rimraf build",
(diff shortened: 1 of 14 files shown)


robhuzzey:
Looks good to me! Thanks for doing this :) :+1:

hxpaul:
I also have need of this, if I've understood it correctly...

mridgway:
lgtm

robhuzzey:
Thanks for this!

Will this be tagged as a release?

PR #197, opened 2015-10-30
Hoist static methods of wrapped component in `injectIntl`

shriah:
This hoists the static method of the wrapped component in `injectIntl` HOC.
ref #196

Issue #196, opened 2015-10-30
static methods not hoisted by `injectIntl`


shriah:
When I tried using `injectIntl` I was not able to access the static method of the child component. The `connect` component of [react-redux](https://github.com/rackt/react-redux/blob/084cedbde87dd88f8a97ac703624ab021c69c355/src/components/connect.js#L6) is using [hoist-non-react-statics](https://github.com/mridgway/hoist-non-react-statics) for the same purpose. May be doing the same with `injectIntl` will be helpful

ericf:
For the time being you can use `hoist-non-react-statics` in your app code to get around the issue. I'm curious to see the actual real world example where you have a component using another component's static memebers. Could you show me so I have a better idea?

shriah:
I am using a static method this along with redux-router to make api calls and wait for them before rendering the components in server side. [Example](https://github.com/erikras/react-redux-universal-hot-example/blob/47fbd7431b1cc0f26a760257c0aabbe20f95123a/src/containers/App/App.js)

ericf:
@shriah I spent some time digging into this, and I'm not sure why the `fetchData()` pattern is using a react component's statics for this. Also, it looks like react-router is [moving away from using statics](https://github.com/rackt/react-router/blob/master/UPGRADE_GUIDE.md#willtransitionto-and-willtransitionfrom).

I'm thinking this simply a limitation of CommonJS, since you have to hang everything off the "default" export in CommonJS, people want the main export to be the component and therefore have to hang things off the constructor. With ES6 modules you can export the component as the `default` export and a function like `fetchData()` as a named export.

With ES6 modules one would:

```js
export default injectIntl(Component);
export function fetchData() {}
```

shriah:
@ericf  `fetchData()` is a user defined function and not related to react-router so this will not be a problem.

The problem with exporting `fetchData` is that, I have to import all the `fetchData` from all my components in my server side code to initiate the api calls. With statics I don't have to worry about importing it and it is already there for my use. For each request `fetchData` call will be made on different components depending on the requested page.

I also feel using HOC should not restrict the functionality of the wrapped component.

I am not sure there is a better way to do this without static functions

``` js
Promise.all(
      
      // Get array of components that will be rendered from the router state.(Not all components)
      // The list of components will be different for each request
      initialState.router.components
          // We need only components with `fetchData` method
          .filter(component => component.fetchData)
          // Fetching the data!
          // `args` is object with required arguments for `fetchData` method
[...]

PR #219, opened 2015-11-16
Allow introspection access to the component wrapped by InjectIntl

akx:
Access to the wrapped components (when not mounted) is useful for universal apps; for instance [@erikras' universal example](https://github.com/erikras/react-redux-universal-hot-example/blob/61610b154fa82d274d89f6cd1f0818edce5a8f6c/src/helpers/getDataDependencies.js) depends on being able to find certain static functions.

This PR also fixes an incidental typo in `injectIntl` and a couple mis-copypasted tests.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -55,5 +55,7 @@ export default function injectIntl(WrappedComponent, options = {}) {
         intl: intlShape,
     };
 
+    InjectIntl.WrappedComponent = WrappedComponent;
+
     return InjectIntl;
 }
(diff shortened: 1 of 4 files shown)


ericf:
@akx thanks for finding and fixing my silly typos! As for the adding the `WrappedComponent` property, this seems related to #196 and #197. Do you mind adding to the discussion in #196? I'm trying to have a full understanding of what people are looking to do with this feature.

analyst74:
@ericf just to add context, allowing inspecting into the wrapped component allows unit testing with shallow rendering. But in that use case, hoisting is not necessary, and being able to access WrappedComponent like in this PR is sufficient.

PR #222, opened 2015-11-17
Add postinstall script, allow install from git

hxpaul:
If you add a post install step we can require this from the git repo too


ericf:
This causes double builds when installing locally. Also this will cause the build to run when anyone installs the package, even from npm.

PR #230, opened 2015-11-24
Upgrade to Babel 6

ericf:
- [x] Update React Intl build to use Babel 6
- [x] Update React Intl examples' builds to use Babel 6
- [x] Improve dist build (smaller, faster)
- [x] Release `babel-plugin-react-intl@2.0.0`


test/setup/babel-register.js
```diff
@@ -0,0 +1,6 @@
+require('babel-core/register')({
+    presets: [
+        'es2015',
+        'react',
+    ],
+});
```
evocateur: This is a really interesting approach to providing the `presets` config to tests, and yet avoiding it in the "global" root `.babelrc`. Why wasn't repo-global config possible?

package.json
```diff
@@ -67,15 +76,13 @@
...
-    "build:dist:main": "browserify src/react-intl.js -s ReactIntl -g [envify --NODE_ENV development] -t babelify -t browserify-shim -d | exorcist -b ./ -r react-intl:/// dist/react-intl.js.map > dist/react-intl.js",
-    "build:dist:main:min": "browserify src/react-intl.js -s ReactIntl -g [envify --NODE_ENV production] -t babelify -t browserify-shim | uglifyjs -c -m > dist/react-intl.min.js",
-    "build:dist:with-data": "browserify src/react-intl-with-locales.js -s ReactIntl -g [envify --NODE_ENV development] -t babelify -t browserify-shim -d | exorcist -b ./ -r react-intl:/// dist/react-intl-with-locales.js.map > dist/react-intl-with-locales.js",
-    "build:dist:with-data:min": "browserify src/react-intl-with-locales.js -s ReactIntl -g [envify --NODE_ENV production] -t babelify -t browserify-shim | uglifyjs -c -m > dist/react-intl-with-locales.min.js",
-    "build:dist": "npm run build:dist:main && npm run build:dist:main:min && npm run build:dist:with-data && npm run build:dist:with-data:min",
+    "build:data": "babel-node --presets es2015 scripts/build-locale-data",
+    "build:lib": "babel --presets es2015,react src/ --ignore src/locale-data/ --out-dir lib/",
```
baer: Why is one `babel-node` and the other just `babel`?. Also, shouldn't presets be handled by the `.babelrc` file?

package.json
```diff
@@ -67,15 +76,13 @@
...
+    "build:dist:dev": "babel-node --presets es2015 scripts/build-dist",
+    "build:dist:min": "NODE_ENV=production babel-node --presets es2015 scripts/build-dist",
```
baer: Aren't presets be handled by the `.babelrc` file?

package.json
```diff
@@ -60,6 +64,11 @@
     "react-addons-test-utils": "^0.14.2",
     "react-dom": "^0.14.2",
     "rimraf": "^2.4.2",
+    "rollup": "^0.21.0",
```
baer: Interesting choice :). Webpack is the standard but Rollup does some really neat stuff. The future is now!

scripts/build-dist.js
```diff
@@ -0,0 +1,92 @@
+import * as p from 'path';
+import * as fs from 'fs';
+import rollup from 'rollup';
+import babel from 'rollup-plugin-babel';
+import npm from 'rollup-plugin-npm';
+import commonjs from 'rollup-plugin-commonjs';
+import replace from 'rollup-plugin-replace';
+import UglifyJS from 'uglify-js';
```
baer: Shouldn't this be lowercased?

scripts/build-dist.js
```diff
@@ -0,0 +1,92 @@
...
+                jsnext: true,
+                skip: [
+                    'react',
+                ],
+            }),
+            commonjs({
+                sourceMap: true,
+            }),
+            replace({
+                'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
```
baer: No dev build?

scripts/build-dist.js
```diff
@@ -0,0 +1,92 @@
...
+    console.log(`Writing: ${p.relative('.', `${dest}.map`)}`);
+}
+
+// -----------------------------------------------------------------------------
+
+createBundle()
+    .then((bundle) => {
+        writeBundle(bundle, {minify: process.env.NODE_ENV === 'production'});
+    })
+    .catch((err) => console.error(err));
```
baer: Why not just throw?

baer:
This looks awesome! Nice work, as usual. :muscle:

test/setup/babel-register.js
ericf: The rollup-plugin-babel package has its own `es2015-babel` preset which conflicts if you also try to use the standard `es2015` Babel preset because Babel merges the collection of presets.

That said, I should probably explicitly list all of the ES2015+ syntax I want to use in the library by _not_ using presets and instead using all `plugins`.

package.json
ericf: This is just using the Babel CLI directly, while the others are scripts I wrote using ES2015 syntax and using `babel-node` to compile and run those custom scripts.

package.json
ericf: The reason I couldn't include them in the `.babelrc` file is that they are merged and Rollup uses its own custom `es2015-rollup` preset: https://github.com/yahoo/react-intl/pull/230#discussion_r45945267

package.json
ericf: This library originally used ES6 Module Transpiler, which changed to Esperanto, which is now Rollup. I'm all for the concept of module "folding" for building library bundles.

scripts/build-dist.js
ericf: I just based it on the README where it had examples of using the Uglify API.

scripts/build-dist.js
ericf: There's a dev build. The default value is `undefined`, I can explicitly set it to `"development"` in the `package.json` `scripts`.

scripts/build-dist.js
ericf: During my testing the error was being swallowed, so I just had it log directly to the console.

ericf:
@evocateur @baer thanks for looking this over! I've tweaked a few things with the Babel config and `scripts` in `package.json`. Hopefully things are now more clearer and more idiomatic.

package.json
```diff
@@ -37,13 +37,17 @@
...
-    "browserify": "^11.0.1",
-    "browserify-shim": "^3.8.10",
-    "envify": "^3.4.0",
+    "babel-plugin-react-intl": "^2.0.0",
+    "babel-plugin-transform-object-rest-spread": "^6.1.18",
+    "babel-preset-es2015": "^6.1.18",
+    "babel-preset-es2015-rollup": "^1.0.0",
+    "babel-preset-react": "^6.1.18",
+    "babelify": "^7.2.0",
+    "browserify": "^12.0.1",
```
baer: Can't both of these + the shim get removed now that Rollup is rolled in? :point_left: see what I did there :)

package.json
```diff
@@ -67,15 +76,13 @@
...
-    "build:dist:main:min": "browserify src/react-intl.js -s ReactIntl -g [envify --NODE_ENV production] -t babelify -t browserify-shim | uglifyjs -c -m > dist/react-intl.min.js",
-    "build:dist:with-data": "browserify src/react-intl-with-locales.js -s ReactIntl -g [envify --NODE_ENV development] -t babelify -t browserify-shim -d | exorcist -b ./ -r react-intl:/// dist/react-intl-with-locales.js.map > dist/react-intl-with-locales.js",
-    "build:dist:with-data:min": "browserify src/react-intl-with-locales.js -s ReactIntl -g [envify --NODE_ENV production] -t babelify -t browserify-shim | uglifyjs -c -m > dist/react-intl-with-locales.min.js",
-    "build:dist": "npm run build:dist:main && npm run build:dist:main:min && npm run build:dist:with-data && npm run build:dist:with-data:min",
+    "build:dist:dev": "NODE_ENV=development babel-node scripts/build-dist",
+    "build:dist:prod": "NODE_ENV=production babel-node scripts/build-dist",
+    "build:dist": "npm run build:dist:dev && npm run build:dist:prod",
```
baer: Why are you building the prod AND dis bundle. Maybe I missed it but it looks like the only difference between the two will be a minified vs a non-minified file.

scripts/build-dist.js
baer: It's not swallowed exactly :smile:. It's just that Promises being a bit more functional in their design don't throw synchronous exceptions. Have a look at [this gist](https://gist.github.com/benjamingr/0237932cee84712951a2
) for a good explanation and a more general fix (hopefully) to the problem. For a much more detailed answer to why this happens and what's being done to relieve this issue youc an check out  [Domenic's What-WG proposal](https://lists.w3.org/Archives/Public/public-whatwg-archive/2014Sep/0024.html).

ericf: Yeah I knew it was a top-level Promise rejection handling issue, but I didn't know that the `unhandledRejection` event existed. I can switch to using that.

package.json
ericf: The development build contains many warns which are removed in the production build.

scripts/build-dist.js
baer: Well, IMHO you may want to consider at least re-throw after you log it so that if this script gets refactored down the road to be something that returns that promise that the error propagates out since what you have hear actually does swallow the err.

Also, I wanted to make sure to toss some resources in a comment just in case somebody finds this PR and thinks (as a lot of people do) that Promises swallow errors :)

package.json
baer: Ah, yea. I missed this line https://github.com/yahoo/react-intl/pull/230/files#diff-56889bb697a98f76cf10bd50d248e636R50

It wasn't clear to me that the dev build and the prd build produced different files. I thought that one would just overwrite the other. Makes sense. Sorry about the spam.

PR #233, opened 2015-12-01
Avoid expensive formatting for simple messages

ericf:
This avoid expensive message formatting for simple messages without `values`. In development messages will always be formatted in case of missing `values` that are suppose to be there.

**This should drastically improve performance, both memory and CPU, for all apps.**

### DIFF
--- a/src/format.js
+++ b/src/format.js
@@ -131,7 +131,14 @@ export function formatMessage(config, state, messageDescriptor, values = {}) {
     // `id` is a required field of a Message Descriptor.
     invariant(id, '[React Intl] An `id` must be provided to format a message.');
 
-    const message = messages && messages[id];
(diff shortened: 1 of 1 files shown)


kfay:
This is going to help a ton! Thanks @ericf. :+1:

caridy:
looks good but maybe we should just wire this up at the beginning of the method, not need to skip a bunch of conditions if no values are provided, and we are in prod. from the current diff it is difficult to imagine that this is a short circuit process, you have to read the whole thing to encounter that it reality it is just falling back to message.

ericf:
@caridy yeah the early return makes the diff much easier to read. I updated to use that instead.

caridy:
:8ball:

PR #242, opened 2015-12-08
Check for preset locale data in hasLocaleData()

caccialdo:
For projects that do not want to use `react-intl`'s addLocaleData method and instead rely on IntlMessageFormat & IntlRelativeFormat own methods.

This PR adds an extra check to see if the locale has already been set on both libs in case it is not found in the registeredLocales hash.

### DIFF
--- a/src/locale-data-registry.js
+++ b/src/locale-data-registry.js
@@ -22,5 +22,6 @@ export function addLocaleData(data = []) {
 }
 
 export function hasLocaleData(locale) {
-    return !!registeredLocales[locale && locale.toLowerCase()];
+    locale = locale && locale.toLowerCase();
+    return !!registeredLocales[locale] || !!(IntlMessageFormat.__localeData__[locale] && IntlRelativeFormat.__localeData__[locale]);
 }


ericf:
Thanks! I tweaked this to remove the locale registry for this package and always delegate to `IntlMessageFormat`'s and `IntlRelativeFormat`'s.

caccialdo:
:+1:

ericf:
@caccialdo I'm working providing better locale fallbacks when data isn't loaded for a particular locale. Naturally I'd want need track of the locale hierarchy based on what comes into React Intl's `addLocaleData()` function. With this change, it complicates the code for looking up the parent locale. I'm curious why you don't want to use React Intl's `addLocaleData()` function. Could you explain the reasoning so I can better understand?

caccialdo:
After starting several projects, one using React, others Ember we realised we could prepare bundles for each locale that:
* Load IntlPolyfill on window
* Replace window.Int by the polyfill (we noticed some discrepancies between V8's Intl and the polyfill and prefer consistency over bundle weight)
* Load IntlMessageFormat on window
* Load the locale data for the given locale
* Load IntlRelativeFormat on window
* Load the locale data for the given locale

These bundle are living on a CDN. They are loaded in the app as part of a separate script tag and webpack is told to look on the window object when requiring one of these 3 packages. This way the main app doesn't have to setup any locale since it's all done for it in the bundle.

By using the lower level message format and relative format addLocaleData API, we stay framework agnostic and I'd prefer so stay like this.

I realise I could prepare similar bundles for react-intl that replace the last 4 points by:
* Load ReactIntl on the window object
* Load locale data using its addLocaleData
and tell webpack to look on the window object when requiring react-intl but that would make these bundles less reusable.

This is the reason I'd prefer to avoid using an `addLocaleData` method tied to a particular framework but I could be convinced otherwise.

ericf:
Okay, this is helpful, thanks! I want to support your approach, which is a more advanced setup than most have. React Intl's `addLocaleData()` is merely intended to be a convenience so people don't have to know about the details of the `IntlMessageFormat` and `IntlRelativeFormat` lower-level libraries.

This change where will go out with the next release. Thanks!

PR #246, opened 2015-12-11
Correct the CLDR acronym

mattparrilla:
Pretty insignificant. Might save someone time searching.

The "Common Locale Data Repository" acronym was incorrectly represented as CDLR, instead of CLDR


ericf:
Thanks!

PR #247, opened 2015-12-11
Add async example

thejameskyle:
My team had some confusion while setting it up this way, hopefully this example will clear up any confusion people have in the future


ericf:
Thanks!

PR #255, opened 2015-12-22
Add styling to FormattedRelative

nodkz:
Add ability to pass styles and classname to FormattedRelative via props.

PR #257, opened 2015-12-22
Add styling to component FormattedMessage

nodkz:
Add ability to pass `style`s and `className`s to FormattedMessage via props.


ericf:

PR #260, opened 2015-12-22
Add ability to pass styles and classNames to FormattedPlural via props.

nodkz:
Also in components renamed property `style` to `intlStyle`.
Property name `style` should be used for decorating elements via css rules.

Internaly `intlStyle` property renamed to `style` when component props passing to internalization methods.


ericf:

PR #261, opened 2015-12-22
Add styling to FormattedTime

nodkz:
Add ability to pass `style`s and `className`s to FormattedTime via props.


ericf:

PR #282, opened 2016-01-13
Refactor build and package layout

### LINKED ISSUE #203: Cannot resolve module 'react-intl/locale-data/ru'
Webpack cannot resolve any `locale-data`

```
Module not found: Error: Cannot resolve module 'react-intl/locale-data/ru' in G:\work\webapp\code\client
resolve module react-intl/locale-data/ru in G:\work\webapp\code\client
  looking for modules in G:\work\webapp\node_modules
    resolve 'file' or 'directory' locale-data/ru in G:\work\webapp\node_modules\react-intl
      resolve file
        G:\work\webapp\node_modules\react-intl\locale-data\ru doesn't exist
        G:\work\webapp\node_modules\react-intl\locale-data\ru.json doesn't exist
        G:\work\webapp\node_modules\react-intl\locale-data\ru.js doesn't exist
      resolve directory
        G:\work\webapp\node_modules\react-intl\locale-data\ru doesn't exist (directory default file)
        G:\work\webapp\node_modules\react-intl\locale-data\ru\package.json doesn't exist (directory description file)
```

The reason is that there's no such folder.

The correct way would be `require('react-intl/lib/locale-data/ru')`.
You should update your documentation.

### LINKED ISSUE #204: Missing locale data for: "en-US". Using default locale: "en" as fallback.
```
[React Intl] Missing locale data for: "en-US". Using default locale: "en" as fallback.
```

### LINKED ISSUE #275: how to just import zh-Hant?
I already setup my own translation, but seems I must also add locale-data from react-intl

so I added

```
import en from 'react-intl/lib/locale-data/zh';
import zh from 'react-intl/lib/locale-data/zh';
addLocaleData(en);
addLocaleData(zh);
```

but zh returning non tranditional one, I saw react-intl/lib/locale-data/index.js contains zh-Hant

how can I add that?

### LINKED ISSUE #277: Support "react-intl/locale-data/*" module imports
It should be possible to have code that looks like the following and have it work in Node.js and when bundled for the browser using Browserify/Rollup/Webpack:

```js
import {addLocaleData} from 'react-intl';
import english from 'react-intl/locale-data/en';
import spanish from 'react-intl/locale-data/es';

addLocaleData(english);
addLocaleData(spanish);
```

According to #249, these module names don't appear to resolve. This is likely because `main` in `package.json` is set to `./lib/react-intl.js` instead of `./lib/index.js`. I'm unsure how the `browser` field in `package.json` will affect the resolution of `'react-intl/locale-data/*'` when bundling with Browserify/Rollup/Webpack… so that'll need to be tested.

ericf:
This is a major refactor of React Intl's build and npm package layout.

Locale data is now package-relative and can more easily be loaded:
```js
import enLocaleData from 'react-intl/locale-data/en';
```

There is now only one copy of these locale data files in the package and they are UMD modules which export the locale data for a language.

Rollup is now being used to build a _single-file_ React Intl module for three different targets:

- **UMD (`ReactIntl` global):** Browser build with bundled dependencies
- **CommonJS (`main`):** For Node/Browserify/Webpack with `require()`-d dependencies
- **ES6 (`jsnext:main`):** For consumption via `import` with `import`-d dependencies

All of the builds are compiled with Babel and bundled into a single module file. Rollup is used to fold all of the internal `src/` modules into one. The difference between `lib/index.js` and `lib/index.es.js` is the latter has everything compiled expect the `import`/`export` statements for React Intl as a whole package.

The `.babelrc` files have been moved to closer to the source code which they apply. This allows for test or example code to leverage experimental ECMAScript syntax without leaking it into the main src.

These changes should result is a smaller and faster package as all of React Intl's code is folded into a single module scope, and the Babel helpers are only included once (instead of per module file). As well as fixing support for IE8.

Fixes #203
Fixes #204
Fixes #275
Fixes #277


SimenB:
Sweet! A timeline for the release of a IE8-compatible version? :smile:

ericf:
@SimenB blocked by this: https://github.com/rollup/rollup-plugin-babel/pull/34

PR #297, opened 2016-01-21
Integration merging static properties in injectIntl

mass3ff3ct:
Hi, I solved the problem of maintaining the static properties of the component after the decoration.

Problem:
```
import * as React from "react"
import {injectIntl} from "react-intl"

@injectIntl
class MyComponent extends React.Component {
     static myProperty = "abc"
     ....
}

console.log(MyComponent.myProperty) // not exists
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ lib/
 locale-data/
 node_modules/
 src/en.js
+.idea/
+npm-debug.log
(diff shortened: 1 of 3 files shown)


ericf:
This has already been discussed. Please check out the several other issue on this topic.

Strate:
#196

PR #300, opened 2016-01-25
Pre-emptively remove 0.14 peerDep

iamdustan:
React 0.15’s first alpha release has been published. This is an effort to avoid requiring all consumers to hold off on updating or suffer through npm’s warn peer dep mismatch message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,9 +36,6 @@
     "intl-relativeformat": "^1.2.0",
     "invariant": "^2.1.1"
   },
-  "peerDependencies": {
-    "react": "^0.14.0"
-  },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


ericf:
The plan will be to support both versions of React as long as there's no changes that would break something fundamental to React Intl.

iamdustan:
That’s what this PR is meant to support. Currently, if I install React’s 0.15-alpha.1 alongside the latest React-Intl beta release NPM 2 gets whiny about the peer dep mismatch.

ericf:
I will test with React 0.15-alpha.1 and if it works I'll add it to the peerDeps list. But I don't want to remove the peerDeps list entirely.

PR #309, opened 2016-02-04
Update README.md from the translations example

alejandroiglesias:
Missing `npm install` before `npm run build`.


ericf:
It's not required, all the deps are included at the root `package.json`.

alejandroiglesias:
I see, thanks for the notice.

PR #328, opened 2016-02-24
add tagName prop handling to component FormattedRelative

johannestroeger:
- add tagName prop handling
- drop React use createElement instead
- add test


ericf:
See the other issues/PRs on this topic.

PR #353, opened 2016-03-09
React 15

ericf:
This adds `react@^15.0.0-rc.1` to `peerDependencies` and updates scripts to test with React 14 and 15.


mridgway:
:+1:

PR #354, opened 2016-03-09
Add React Router example

ericf:


baer:
This is a great example! Thanks for writing this. I haven't had time to do the 1.x analysis but when I do I'll see if I can either make a PR with an example or just a comment in the docs.

ericf:
@baer cool, sounds good!

anomaly44:
@baer can you provide more info, I m having the same problem using SSR, but I'm clueless on where to look. Thanks in advance.

MasonChinkin:
> @baer can you provide more info, I m having the same problem using SSR, but I'm clueless on where to look. Thanks in advance.

@anomaly44 I'm having the same problem as you 3 years in the future after upgrading to React 16 SSR. Did you recommend any material?

anomaly44:
@MasonChinkin I can't remember what it was exactly. But make sure to load the correct locale en make sure all needed props are there on the server. Havent used a custom SSR method anymore in the last couple of years. So that's all I can say at the moment.

MasonChinkin:
@anomaly44 thanks. Turned out that while the error message was pointing to intl and react-router, the problem was actually a version of react-transition-group incompatible with react 16. Fun stuff...

PR #358, opened 2016-03-10
Use specific thresholds for relative formatting

ericf:
The `intl-relativeformat` package has a set of thresholds to round-up to the next significant unit that make sense for static formatting. But these thresholds don't make sense when `<FormattedRelative>` automatically "ticks" the relative times.

This change adjusts the thresholds to not round-up, and does so in a way that shouldn't effect other uses of `intl-relativeformat` because the thresholds are only temporarily updated, then put back after formatting.

### DIFF
--- a/src/format.js
+++ b/src/format.js
@@ -5,6 +5,7 @@
  */
 
 import invariant from 'invariant';
+import IntlRelativeFormat from 'intl-relativeformat';
 
 import {
     dateTimeFormatPropTypes,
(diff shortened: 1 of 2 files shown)


src/format.js
```diff
@@ -96,6 +97,17 @@ export function formatRelative(config, state, value, options = {}) {
...
     let filteredOptions = filterProps(options, RELATIVE_FORMAT_OPTIONS, defaults);
 
+    // Capture the current threshold values, then temporarily override them with
+    // specific values just for this render.
+    const thresholds = {...IntlRelativeFormat.thresholds};
+    Object.assign(IntlRelativeFormat.thresholds, {
+        second: 60, // seconds to minute
+        minute: 60, // minutes to hour
+        hour  : 24, // hours to day
+        day   : 30, // days to month
```
ericf: The only question is what the days-to-month value should be. Should it be 28, 29, 30, or 31? With it set at `30`, by default it will go from `"29 days ago"` to `"1 month ago"`.

lingyan: what does moment.js do?

ericf: I'll have to look again, but we took the original thresholds from Moment :) and it has `26` for days-to-month.

ericf: Yeah, it's still `26`: https://github.com/moment/moment/blob/2fb832be8c13d8164f9282826a362057af5cb66f/src/lib/duration/humanize.js#L8

etast:
LGTM. I know see 47 minutes when testing

PR #377, opened 2016-03-29
Update to eslint@2 and eslint-plugin-react@4, fix linting issues

hzoo:
Interesting issue with greenkeeper is that it's possible that only updating 1 of the dependencies will make tests failing (linting) and with multiple it will pass.

2 warnings are TODO comments

cc @ericf


.eslintrc
```diff
@@ -23,7 +23,7 @@
         "indent": [2, 4],
         "linebreak-style": [2, "unix"],
         "prefer-arrow-callback": 2,
-        "quotes": [2, "single"],
+        "quotes": [2, "single", "avoid-escape"],
```
hzoo: Might not be necessary since I just changed the template string to single quotes.

.eslintrc
```diff
@@ -36,6 +36,7 @@
         "no-var": 2,
         "no-warning-comments": 1,
 
-        "react/jsx-uses-react": 2
+        "react/jsx-uses-react": 2,
+        "react/jsx-uses-vars": 2
```
hzoo: This will remove the unused jsx variable issue (most of the errors)

examples/translations/scripts/lib/printer.js
```diff
@@ -23,18 +23,20 @@ export default function printICUMessage(ast) {
 
         let formatType = format.type.replace(/Format$/, '');
 
+        let style, offset, options;
```
hzoo: Otherwise can remove this rule:

`29:9  error  Unexpected lexical declaration in case block  no-case-declarations`

package.json
```diff
@@ -57,8 +57,8 @@
     "browserify": "^13.0.0",
     "browserify-shim": "^3.8.11",
     "es6-promise": "^3.0.2",
-    "eslint": "^1.2.1",
-    "eslint-plugin-react": "^3.2.3",
+    "eslint": "^2.5.3",
+    "eslint-plugin-react": "^4.2.3",
```
hzoo: Update both to next major version

ericf:
:heart: thanks @hzoo!

PR #383, opened 2016-03-31
Remove babel-eslint

ericf:
`babel-eslint` released a patch version which caused the built to break. This removes `babel-eslint` because it's not currently needed.


hzoo:
The patch version had to do with `"sourceType": "module"` since we were hardcoding the value in babel-eslint, but it was reverted since it was breaking. But if you don't need it :+1:

PR #392, opened 2016-04-07
Make time tags HTML 5

LarsOL:
Use the HTML 5 tags, I have the suspicion that this should be  added to Formatted Relative, but since that is a more complicated case I left it to you.

Btw, Since I'm on windows I couldn't get the 'npm i' to work due to the 'NODE_ENV=development' in the scripts failing on windows, so I couldn't check if test fail due to this.

### DIFF
--- a/src/components/date.js
+++ b/src/components/date.js
@@ -28,7 +28,7 @@ export default class FormattedDate extends Component {
             return children(formattedDate);
         }
 
-        return <span>{formattedDate}</span>;
+        return <time datetime={value}>{formattedDate}</time>;
     }
 }
 
--- a/src/components/time.js
+++ b/src/components/time.js
@@ -28,7 +28,7 @@ export default class FormattedTime extends Component {
             return children(formattedTime);
         }
 
-        return <span>{formattedTime}</span>;
+        return <time datetime={value}>{formattedTime}</time>;
     }
 }
 


src/components/date.js
```diff
@@ -28,7 +28,7 @@ export default class FormattedDate extends Component {
             return children(formattedDate);
         }
 
-        return <span>{formattedDate}</span>;
+        return <time datetime={value}>{formattedDate}</time>;
```
ericf: Why not `<date>` here? Also the `datetime` attribute requires a specific format, and for the value to be a string.

LarsOL: <date> is not a valid HTML5 tag?

And yes that is true is there a parsed format of the input date we can easily get to?

LarsOL:
Ill be spinning up a Linux dev environment later in this month. So I can clean up the failing tests and search of a ISO parsed date to put in the datetime. 

Before I start, would you actually want this change? and also the more i think about it, I also believe the time tag should be added to formattedRelative component too, what are your thoughts?

ericf:
I think this is part of a larger change to rethink what level of abstraction the `<Formatted*>` components should be providing. I think the focus can be on accessibility for the end user, and usability for the developer. There's also the concern about perf as this will be more costly to render the component.

One major thing I see with `<FormattedMessage>` is that it's used for display next, but coupled with a `formatMessage()` call for the `aria-label` or `title` attribute; e.g.:

```js
const SomeComponent = injectIntl(({intl}) => (
    <span title={intl.formatMessage({id: 'tooltip'})}>
        <FormattedMessage id='text'/>
    </span>
));
```

This is also a common thing which I think can be improved by a concept of an `<IntlComponent>`, similar to [`<CoreComponent>`](https://github.com/necolas/react-native-web/blob/master/src/components/CoreComponent/index.js) in the React Native Web project.

ericf:
Closing this PR because the build fails and this change should be part of a larger conversation about the role of the `<Formatted*>` components.

PR #395, opened 2016-04-07
Add "postversion" to remove "history" before publish

### LINKED ISSUE #394: `history` packaged is being bundled with npm package
The `react-intl` npm package is bundling the `history` npm package. This is because of an _old_ npm bug:

- https://github.com/reactjs/react-router/issues/2195
- https://github.com/npm/npm/issues/9894
- https://github.com/npm/npm/issues/9642
- https://github.com/npm/npm/pull/11995

This should be fixed by removing the `history` package from `node_modules` before publishing the npm package.

ericf:
This fixes an npm packaging bug with the `history` npm package being bundled.

Fixes #394

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,6 +110,7 @@
     "test:watch": "npm run test -- --watch --reporter min",
     "test:perf": "NODE_ENV=production babel-node test/perf",
     "preversion": "npm run clean && npm run build && npm run test:all",
+    "postversion": "echo 'Uninstall react-router to fix npm issue' && npm uninstall react-router",
     "prepublish": "npm run clean && npm run build"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-intl",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "Internationalize React apps. This library provides React components and an API to format dates, numbers, and strings, including pluralization and handling translations.",
   "keywords": [
     "intl",


package.json
```diff
@@ -110,6 +110,7 @@
     "test:watch": "npm run test -- --watch --reporter min",
     "test:perf": "NODE_ENV=production babel-node test/perf",
     "preversion": "npm run clean && npm run build && npm run test:all",
+    "postversion": "echo 'Uninstall react-router to fix npm issue' && npm uninstall react-router",
```
ericf: This is kind of a hack to do this in "postversion". Ideally, I'd do this in "prepublish", but that would mean that when you clone the repo and run `npm install` locally it will remove the `history` package. During development, or if someone is testing out the examples, I don't want this package the be removed, so that's why I'm doing it on "postversion" where the workflow is:

```
$ npm version patch
$ npm publish
```

taion:
You can also publish with npm@2, if you're so inclined, as this bug doesn't arise in npm@2.

Although my experience with trying to do this so far has mostly consisted of forgetting to downgrade to npm@2 locally, and consequently publishing broken packages.

ericf:
> Although my experience with trying to do this so far has mostly consisted of forgetting to downgrade to npm@2 locally, and consequently publishing broken packages.

@talon yeah this is exactly my worry :)

taion:
This bug is finally fixed for npm3 on npm@next (3.8.7), though: https://github.com/npm/npm/pull/11995

PR #405, opened 2016-04-11
Tweaking README

okuryu:
We don't need to have links to http://formatjs.io/react/ because the
http://formatjs.io/react/ redirects to https://github.com/yahoo/react-intl.

We don't have any mentions about ICU Message syntax in the README.


ericf:
Thanks!

PR #421, opened 2016-04-20
Perf Improvements: shouldIntlComponentUpdate and nested IntlProviders

ericf:
Improved the perf of `shouldIntlComponentUpdate` util which now checks if `context.intl` is the same as `nextContext.intl` before doing a shallow compare of their filtered props.

Creating `Intl*` formatter instances is expensive, so `<IntlProvider>` memoizes the `Intl*` constructors to reuse the formatter instances. These memoized formatters are now shared with nested `<IntlProvider>` instances.

This sharing should increase the rendering perf of component architectures which use nested `<IntlProvider>` instances.

### DIFF
--- a/src/components/provider.js
+++ b/src/components/provider.js
@@ -38,6 +38,8 @@ export default class IntlProvider extends Component {
             'See: http://formatjs.io/guides/runtime-environments/'
         );
 
+        const {intl: intlContext} = context;
+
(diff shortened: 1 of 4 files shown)


mridgway:
:+1:

kfay:
👍

Issue #419, opened 2016-04-19
Passing <FormattedMessage> to a component inside a nested <IntlProvider>


wrwrwr:
With 2-beta I used to import `formatMessage()` and its siblings from `react-intl/lib`. Obviously that was undocumented, unsupported, etc. Nevertheless, it may have been the best approach to [toy with](https://github.com/wrwrwr/react-intl-ns/blob/master/namespaces.jsx#L24) the `config` parameter that the `format*` functions [get bound to](https://github.com/yahoo/react-intl/blob/master/src/components/provider.js#L110) during `IntlProvider` child context construction (I've actually tried a couple of different options with less pleasing outcomes).

Now the only way to reuse these functions seems to be to check out the sources, import from `react-intl/src` and make some adjustments to ones transpiling configuration. Not really an issue, except that it results in increased size of the dependent module -- as it likely means bundling a significant part of react-intl.

That's a pretty obscure case, but how about exporting the `format*` functions, or even better making the unbound versions available from the `config` object (or somewhere else)? Otherwise, just count this as an example for https://github.com/yahoo/react-intl/issues/416.

ericf:
You shouldn't be using the `src/` files, they are only there for the source maps, but maybe should be removed from npm package. Can you explain _why_ you want these functions to be exposed? What's the concrete use case?

wrwrwr:
Yup, I know I should not, nor the former `lib/format` probably :-)

The aim or use case is to introduce namespaces. In short, to allow inserting additional messages and formats scoped to a component (or some other entity), and to let translation components declare which namespace they belong to (thus which message dictionary should be used).

More specifically, the idea is to provide a custom provider that switches message and format dictionaries dynamically, depending on the namespace a given message belongs to.
The provider can be realized by wrapping the `format*` methods living on the `intl` object -- just to pass them a modified `config`. You can't "unbind" the `config` however, hence the original functions are needed.

I believe the first example at [react-intl-ns](https://github.com/wrwrwr/react-intl-ns#readme) may clarify the general design. The only code piece that uses those functions is the one I've linked above.

ericf:
Have you tried nesting `<IntlProvider>`s to accomplish this?

wrwrwr:
[...]

PR #429, opened 2016-04-21
Extended the nested providers example for #419.

wrwrwr:
I've been trying to keep things simple and match the theme, so the example is somewhat artificial :-)
The `Container` and `Message` components could live in separate packages, and know little about namespaces defined or not defined in their parents (passing `getIntlMessages()` down seems to be a bit of unnecessary binding, and `getIntlFormats()` would also be needed). Containers within containers are not uncommon. The "strings" passed down could include some more complex content  (thus `FormattedHTMLMessage` could not be an option).

A more practical scenario would be for the container to be a modal with some textual chrome elements, and instead of a message it could display some canvas / svg chart that would include content passed down from the app.

Technically, we have all elements in a tree, so everything can be done with nested providers. Just wrap each message in the proper provider and pull common providers to ancestors in one way or another. On the other hand, namespaces can be used thoughtlessly and no binding between components needs to be introduced (each component can declare and use its own namespace -- just for its own messages).

### DIFF
--- a/examples/nested/src/client/components/app.js
+++ b/examples/nested/src/client/components/app.js
@@ -1,6 +1,7 @@
 import React, {Component, PropTypes} from 'react';
 import {IntlProvider, FormattedMessage} from 'react-intl';
 import Greeting from './widgets/greeting';
+import Container from './widgets/container';
 
 class App extends Component {
     constructor(props) {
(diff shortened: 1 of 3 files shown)


ericf:
Added some comments back here: https://github.com/yahoo/react-intl/issues/419#issuecomment-213841978

redonkulus:
Closing this as its an example, further discussion can happen in the issue if the feature request is still wanted.

PR #433, opened 2016-04-23
Add hoisting of static methods on components

gustavnikolaj:
Making sure that a component with static methods will have those available after being wrapped with injectIntl.

I had a component with a static method that is used to create an `onEnter` hook in my react-router route definitions. After wrapping that component with `injectIntl` it no longer had that static method.

It was not a problem with react-redux's `connect` component, so I looked at how [it was implementing that](https://github.com/reactjs/react-redux/blob/ea53d6fb076359a864a1d543568d951d4b3eab3d/src/components/connect.js#L365) for reference.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "react": "global:React"
   },
   "dependencies": {
+    "hoist-non-react-statics": "^1.0.5",
     "intl-format-cache": "^2.0.5",
     "intl-messageformat": "^1.3.0",
     "intl-relativeformat": "^1.3.0",
--- a/src/inject.js
+++ b/src/inject.js
@@ -11,6 +11,7 @@ import React, {Component} from 'react';
 import invariant from 'invariant';
 import {intlShape} from './types';
 import {invariantIntlContext} from './utils';
+import hoistStatics from 'hoist-non-react-statics';
 
 function getDisplayName(Component) {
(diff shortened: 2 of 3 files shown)


gustavnikolaj:
I was just made aware of https://github.com/yahoo/react-intl/issues/196 - I'm not sure how I missed that when scanning through issues initially.

I'll leave this PR around though, as I really think it's worth reconsidering.

People use these higher order components like decorators, and the convention that you come to expect as a user of these libraries are that you wrap your Components in the HoCs to augment the Component.

I know that you're right, in saying that they aren't decorators, and that functions like my onEnter hook might make more sense as individual exports - it just that the rest of the react community estabilshed a way of doing things that you chose not to follow. That decision - no matter how right - causes friction. It was really surprising to me that it did not work like I had come to expect.

I hope you'll reconsider.

kellyrmilligan:
just ran into this, it is interesting that injectintl is inspired by connect, yet doesn't hoist statics.

voidale:
+1

StorytellerCZ:
Superseded by #1073

infiman:
Having `preload` static method in my components for my SSR implementation, but `injectIntl` is not hoisting it after wrapping them. Any chance this PR will be merged? Thanks in advance!

PR #434, opened 2016-04-23
FormattedMessage: Pass through DOM props. Closes #432

### LINKED ISSUE #432: Components: pass through dom props
It' very easy to redefine `tagName` with components to render `button` or `a`, but it's impossible to set `onClick` or `href` on it. Even `className` couldn't be set to default `span`. :(
Would be great if components would pass through general dom props to rendered element to avoid extra usage of `injectIntl`.

langpavel:
It allows you to write this JSX:
```jsx
<FormattedMessage {...descriptor} tagName="a" href="http://..." className="css" />
```

### DIFF
--- a/src/components/message.js
+++ b/src/components/message.js
@@ -109,7 +109,14 @@ export default class FormattedMessage extends Component {
             return children(...nodes);
         }
 
-        return createElement(tagName, null, ...nodes);
+        const props = { ...this.props };
+        delete props.id;
+        delete props.description;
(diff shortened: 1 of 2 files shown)


langpavel:
@ericf What do you mean about this? Can you review?

ericf:
https://github.com/yahoo/react-intl/pull/255#issuecomment-166628628

langpavel:
Oh, I see, good luck, there will be much more same issues opened I think..

PR #435, opened 2016-04-25
Properly apply defaultProps to <IntlProvider>

ericf:
Previously if a prop had an `undefined` value, it wouldn't receive the `defaultProps` value. This fixes that.

_Note: `<IntlProvider>` doesn't use React's built-in `defaultProps` resolution because it first inherits for an `<IntlProvider>` in its ancestry._

### DIFF
--- a/src/components/provider.js
+++ b/src/components/provider.js
@@ -79,10 +79,16 @@ export default class IntlProvider extends Component {
 
         // Build a whitelisted config object from `props`, defaults, and
         // `context.intl`, if an <IntlProvider> exists in the ancestry.
-        let config = {
-            ...defaultProps,
-            ...filterProps(this.props, intlConfigPropNames, intlContext),
-        };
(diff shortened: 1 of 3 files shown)


src/components/provider.js
```diff
@@ -79,10 +79,16 @@ export default class IntlProvider extends Component {
...
-        let config = {
-            ...defaultProps,
-            ...filterProps(this.props, intlConfigPropNames, intlContext),
-        };
+        let config = filterProps(this.props, intlConfigPropNames, intlContext);
+
+        // Apply default props. This must be applied last after the props have
+        // been resolved and inherited from any <IntlProvider> in the ancestry.
+        // This matches how React resolves `defaultProps`.
+        for (let propName in defaultProps) {
```
ericf: No need for `hOP()` since I created the `defaultProps` object. React doesn't use it either for the same reason.

carystanley:
👍

PR #441, opened 2016-04-28
Separate delimiter and placeholder for IE8 support

amccloud:
Passing a react element to a placeholder does not work in IE 8. This is because `String.prototype.split` does not include the placeholder in the resulting array.

**Chrome**
```js
'this|is|an|example'.split(/(\|)/g)
['this', '|', 'is', '|', 'an', '|', 'example']
```

**IE 8**
```js
'this|is|an|example'.split(/(\|)/g)
['this', 'is', 'an', 'example']
```

### DIFF
--- a/src/components/message.js
+++ b/src/components/message.js
@@ -49,7 +49,7 @@ export default class FormattedMessage extends Component {
             children,
         } = this.props;
 
-        let tokenRegexp;
+        let tokenDelimiter;
         let tokenizedValues;
         let elements;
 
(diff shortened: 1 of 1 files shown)


yahoocla:

ericf:
@amccloud thanks! Is this the only IE8 bug you've found?

Would you mind signing the Yahoo CLA and adding code comments about IE8 so this doesn't get lost in a future refactor?

yahoocla:
CLA is valid!

amccloud:
This is the only issue in IE8.

Signed CLA and pushed update with comment.

ericf:
Released as `v2.1.2`: https://github.com/yahoo/react-intl/releases/tag/v2.1.2

PR #457, opened 2016-05-12
Added french local

CarlBosch:
extended the translation sample to include a real foreign language (french)

### DIFF
--- a/examples/translations/package.json
+++ b/examples/translations/package.json
@@ -15,7 +15,8 @@
   "license": "BSD-3-Clause",
   "browser": {
     "react-intl": "../../lib/index.js",
-    "react-intl/locale-data/en": "../../locale-data/en.js"
+    "react-intl/locale-data/en": "../../locale-data/en.js",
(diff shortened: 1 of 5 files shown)


yahoocla:

CarlBosch:
agreed on Yahoo CLA

ericf:
@CarlBosch thanks for offering this, but this isn't how the example is suppose to work. The idea is that you only put the default language strings in the source code and the tooling takes care of the rest. For this example to _simulate_ translation, I have a automated script which transforms to upper case.

CarlBosch:
Hi,
I'm quite new to react-intl so please don't mind me asking, was what I proposed a reasonable way to do translation in other languages? Or is there another (better) way to achieve this? 
Thanks,
Carl

ericf:
`defaultMessage` is for defining the default message for the default language. There's `babel-plugin-react-intl` that will do the extraction into a JSON file so that they strings can be provided to a translator.

After having the strings translated, they can then be passed-in at runtime via `<IntlProvider messages={traslatedStringsCollection}>`.

PR #478, opened 2016-06-02
added dev dependency and use of cross-env

cramhead:
I added a devDependency reference to cross-env and adjust the the package.json scripts to use cross-env. I tested it by running ```npm install``` without cross-env it failed. With cross-env it succeeded. I subsequently ran all script where ```NODE_ENV``` was defined with cross-env and the completed with no errors


yahoocla:
CLA is valid!

ericf:
@cramhead thanks so much for making this change and testing it!

PR #488, opened 2016-06-08
Support messages that are empty strings

mieszko4:
I would like to propose allowing of empty messages in `react-intl`.

The reasoning for that would be that for some languages I do not want to display translation and currently I get warnings and `id` in my app instead of empty string.
Other solution for this scenario would be explicitly defining whitelist/blacklist of languages for `<FormattedMessage />` but I feel that I don't want have such logic in the component.

In any case I am not sure what would be the best solution for that, that's why I propose this pr.
If you have any comment on that it would be great to discuss it!


yahoocla:

yahoocla:
CLA is valid!

ericf:
> The reasoning for that would be that for some languages I do not want to display translation and currently I get warnings and id in my app instead of empty string.

Why not use a conditional then? If for some reason you have a blacklist for languages for which you don't want to display a string, then conditionally render that component with that message id.

I'm not in favor of this, because having empty strings in the UI is bad; e.g. an action button with no text inside.

langpavel:
```js
$ node
> !!''
false
> !!' ' 
true
```

@mieszko4 Can one space translation solve your problem?

mieszko4:
@langpavel Yeah, I am using it like that already. To be more specific before using defined messages I add `' '` for all `''`.

@ericf I guess you are right. I did not want to use conditionals cause it seems to me like an overhead in the component code.

But after taking a thought about it I agree that displaying an empty message should be explicit and I cannot think of any example where a translation results with an empty string, thus I will close this pr.

langpavel:
@mieszko4 Eric may want to say that you might be using the `FormattedMessage` wrong.

Did you translate fragments of sentence in multiple strings? This is not good practice. You can inject components into your `message` like this:
```js
const messages = defineMessages({
  ...
  wrapper: {
    id: 'wrapper',
    defaultMessage: 'Click on the {okButton} button or {dismissButton}',
  }
  ...
});
```
You can define conditionals, rearange things in sentence etc..

mieszko4:
Thanx @langpavel.
My use case is simpler than that. Basically I have some sentences in `<p/>` that I do not want to show for specific languages. I understand concern of @ericf about empty button but that is not my use case.

We have `json` translation files that someone else (not a programmer) is modifying. He knows what he wants to hide for some languages and he basically puts `""` in translation files.
We do it that way since it is quicker as no code needs to be modified - so no interaction with programmer.

But I understand it is mixing of concerns.

ericf:
@mieszko4 thanks for providing more context about what you're dealing with.

Like what @langpavel was getting at, could you process the messages collection before passing it to `<IntlProvider>` and replace `""` with `" "`? This would be modeling your business logic.

Another approach would be to create a wrapper component:

```js
import {injectIntl, FromattedMessage} from 'react-intl';

export default injectIntl(({intl, ...props}) => (
    intl.messages[props.id] === '' ? null : <FormattedMessage {...props}/>
));
```

mieszko4:
@ericf I am already using the first approach that you wrote and it works for me very well.

But the wrapper component looks great to me! Thanx for the suggestion!

langpavel:
If you dealing with block of lines, paragraphs, you can create whole string and postprocess it, splitting translated text result by `/\n/` then iterating through `.map(line => <p>line</p>)`

mieszko4:
Thanx @langpavel!

silasb:
@ericf I'm trying to use your solution at https://github.com/yahoo/react-intl/pull/488#issuecomment-224945705 but having a tough issues trying to wrap my head around how you'd use it.

Based on the `export` I'd assume it's a replacement for `injectIntl`, but I'm not sure about this since it's missing key function arguments.

Do you think you could throw a quick example together to show how you'd use this?

ericf:
> Do you think you could throw a quick example together to show how you'd use this?

@silasb here it is re-written in ES5:

```js
var ReactIntl = require('react-intl');

var injectIntl       = ReactIntl.injectIntl;
var FormattedMessage = ReactIntl.FormattedMessage;

function MyFormattedMessage(props) {
    var intl = props.intl;
    return intl.messages[props.id] === '' ? null : <FormattedMessage {...props}/>
}

module.exports = injectIntl(MyFormattedMessage);
```

TikiTDO:
@ericf I have a situation where your solution is not sufficient. 

I have an element that displays:

1. Some translated text, if a raw value is valid
2. An error message if a raw value is not valid
3. Empty text for certain situations where that raw value is valid, but is only used to set a background color, and does not need any text.

Being able to pass an empty string or a space would trivially solve this problem. If a blank default message is such a problem by itself, would it be possible to just add an extra key into the message definition to opt out of the trim behavior?

PR #496, opened 2016-06-14
Fix <FormattedRelative> to update when value changes

### LINKED ISSUE #495: FormattedRelative doesn't update state.now on props.value change.
I'm using react-intl for an app with lots of live data. FormattedRelative is used as a "Last change for this item" marker. When that timestamp changes, FormattedRelative will sometimes display things like "in 27 seconds" as if the last change happened in the future. This of course is because the comparison inside the component is between `this.state.now` and `this.props.value`. The props change happens in between the timer ticks to update `this.state.now`. After the next tick the output is correct again. Check this [fiddle](http://jsfiddle.net/ybm69j1z/2/), click the button and you'll see "in 5 seconds" etc.

One idea is to update `this.state.now` when the `value` prop changes, something like
```javascript
function componentWillReceiveProps(next) {
	if (this.props.value && this.props.value.getTime() != next.value.getTime()) {
		this.setState({now: this.context.intl.now()})
	}
}
```
But you probably have better ideas on how to handle this, or if it shouldn't be handled at all and I should restructure my app somehow instead.

ericf:
This fixes the issue where `<FormattedRelative>` would render the incorrect relative time when its `props.value` was updated.

Fixes #495

### DIFF
--- a/test/unit/components/relative.js
+++ b/test/unit/components/relative.js
@@ -210,6 +210,29 @@ describe('<FormattedRelative>', () => {
         }, 10);
     });
 
+    it('updates when the `value` prop changes', () => {
+        const {intl} = intlProvider.getChildContext();
+        const now = intl.now();
+
+        renderer.render(<FormattedRelative value={now} updateInterval={1} />, {intl});
+        const renderedOne = renderer.getRenderOutput();
+
+        // Shallow Renderer doesn't call `componentDidMount()`. This forces the
+        // scheduler to schedule an update based on the `updateInterval`.
+        renderer.getMountedInstance().componentDidMount();
+
+        // Update `now()` to act like the <IntlProvider> is mounted.
+        const nextNow = now + 1000;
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

test/unit/components/relative.js
```diff
@@ -210,6 +210,29 @@ describe('<FormattedRelative>', () => {
...
+        renderer.getMountedInstance().componentDidMount();
+
+        // Update `now()` to act like the <IntlProvider> is mounted.
+        const nextNow = now + 1000;
+        intl.now = () => nextNow;
+
+        renderer.render(<FormattedRelative value={nextNow} updateInterval={1} />, {intl});
+        const renderedTwo = renderer.getRenderOutput();
+
+        expect(renderedTwo).toEqualJSX(renderedOne);
```
ericf: This assertion failed before the changes in `src/components/relative.js`

PR #508, opened 2016-06-25
Added more tests for pluralization in FormatMessage

cramhead:
I added a few new tests to test the pluralization in FormatMessage. 
Also, the source code has been reformatted with vs code.


yahoocla:
CLA is valid!

PR #509, opened 2016-06-27
The component passed to `injectIntl` should return a component with the same propTypes

gabehayes:
I came across a scenario in which the propTypes were unavailable on the object returned by the `injectIntl` method. For example:

```
export class Rocket extends Component {
  ...
}

Rocket.propTypes = {
  fuel: PropTypes.string,
};

export default injectIntl(Rocket);
```
And then elsewhere:
```
import Rocket from './rocket';

export class Falcon extends Component {
  ...
}

Falcon.propTypes = {
  fuel: Rocket.propTypes.fuel, // Rocket.propTypes is `undefined`
  revision: PropTypes.number.isRequired,
};
```

I realize that they are available on the `WrappedComponent` property, but found this to be undesirable. 
```
Falcon.propTypes = {
  fuel: Rocket.WrappedComponent.propTypes.fuel,
  revision: PropTypes.number.isRequired,
};
```

Another solution was to access the `propTypes` on a named export, which also felt undesirable:
```
export Rocket, { Rocket as NamedRocket } from './rocket';

Falcon.propTypes = {
  fuel: NamedRocket.propTypes.fuel,
  revision: PropTypes.number.isRequired,
}
```

Passing my component to the `injectIntl` should enhance my component, but losing access to `propTypes` directly felt like a degradation to the component.

These proposed changes return a component whose propTypes match those that you originally defined on the component you are enhancing, and also attach the required prop defined by `intlPropName`.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -55,6 +55,11 @@ export default function injectIntl(WrappedComponent, options = {}) {
         intl: intlShape,
     };
 
+    InjectIntl.propTypes = {
+        ...WrappedComponent.propTypes,
+        [intlPropName]: intlShape,
+    };
+
     InjectIntl.WrappedComponent = WrappedComponent;
 
     return InjectIntl;
--- a/test/unit/inject.js
+++ b/test/unit/inject.js
@@ -63,6 +63,20 @@ describe('injectIntl()', () => {
         expect(renderer.getRenderOutput()).toEqualJSX(<Wrapped foo="foo" intl={intl} />);
     });
 
+    it('adds all props from <WrappedComponent> to <InjectIntl>', () => {
+        Wrapped.propTypes = {
+            ...Wrapped.propTypes,
+            foo: React.PropTypes.string,
+            bar: React.PropTypes.string.isRequired,
+        };
+
+        const Injected = injectIntl(Wrapped);
+        const {intl} = intlProvider.getChildContext();
+
+        renderer.render(<Injected foo="foo" bar="bar" />, {intl});
+        expect(Injected.propTypes).toEqual(Wrapped.propTypes);
+    });
+
     describe('options', () => {
         describe('intlPropName', () => {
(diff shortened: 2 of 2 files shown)


yahoocla:
CLA is valid!

alexshepard:
:+1:

victuoso:
👍

ericf:
A solution you can use today is to simply add `.propTypes` to the component returned from `injectIntl()`.

ericf:
@gabehayes thanks for taking the time to create this.

Thinking about this more, I don't think this is the right solution. You can even see in this repo I have a `types.js` file which provides the shared `propTypes` schema for components to use — this is a pattern I've seen in apps use as well.

I don't think `propTypes` should be proxied by the component returned from `injectIntl` because it's metadata that could easily be written another way; e.g. TypeScript or Flow. In fact, there are signals for the React core team that [Flow](https://flowtype.org/docs/react.html) could become the preferred way to annotate types instead of using `propTypes`.

I think you'll find your types to be more composable if you move them out of the modules providing components since they define you're data model which many components may render. I think this repo actually has a good examples of doing this:

- [`dateTimeFormatPropTypes`](https://github.com/yahoo/react-intl/blob/master/src/types.js#L43-L59) from `types.js`
- [`<FormattedDate> propTypes`](https://github.com/yahoo/react-intl/blob/master/src/components/date.js#L41-L46) from `components/date.js`
- [`<FormattedTime> propTypes`](https://github.com/yahoo/react-intl/blob/master/src/components/time.js#L41-L46) from `components/time.js`

gabehayes:
@ericf Thanks for reviewing this. After much thought, I will also concede that mutating the `propTypes` of the new component returned by `injectIntl` does not seem like the right thing to do.

I think the best thing to do in the meantime, in lieu of exploring exportable shapes, is to just access the component directly via the `WrappedComponent` property.

ericf:
@gabehayes or move the types to its own module that components can share. Then you alleviate an owner and refactoring problems.

PR #547, opened 2016-07-26
Produce a nice error if a ReactElement is passed to formatMessage()

STRML:
We monkey-patch this in our development environment. It's useful for e.g. reusing constant elements in multiple places in the code:

```jsx
const CommonMessages = {
  'Error': (<FormattedMessage 
    id="common.error" 
    description="Common error message."
    defaultMessage="Error" />)
};

// ...

render() {
  // should be passing `.props`, not the element - common mistake
  <span title={this.context.intl.formatMessage(CommonMessages['Error'])}>Error</span>
}
```

This produces the error `'[React Intl] An `id` must be provided to format a message.'` which could be more helpful.

This patch produces the error: `'[React Intl] Don\'t pass React elements to formatMessage(), pass `.props`.'` which is more descriptive.

### DIFF
--- a/src/format.js
+++ b/src/format.js
@@ -6,6 +6,7 @@
 
 import invariant from 'invariant';
 import IntlRelativeFormat from 'intl-relativeformat';
+import {isValidElement} from 'react';
 
 import {
     dateTimeFormatPropTypes,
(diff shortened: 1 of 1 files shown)


yahoocla:

STRML:
CLA signed.

yahoocla:
CLA is valid!

ericf:
@STRML thanks for this. I think you could update the `invariant` check that's already there for the `id` check to improve the message when when `messageDescriptor` is a React element.

STRML:
Well, that's essentially what I'm doing.

I don't want to include this check in production - it's relatively more expensive - and it's a separate invariant from simply missing the `id`.

redonkulus:
@STRML looks at old PRs, is this something you still need? If not, please feel free to close.

STRML:
Nothing has changed since the original submission, I think this is still good to merge.

redonkulus:
@STRML ok, can you resolve the conflict?

STRML:
Yeah, done @redonkulus

PR #594, opened 2016-08-15
Upgrade devDeps and pin rollup@0.34.3

ericf:
See: https://github.com/rollup/rollup/pull/823#issuecomment-239867292


yahoocla:
CLA is valid!

PR #597, opened 2016-08-17
Add default values to formatTime

budde377:
Currently, when using `formatTime` and other date-format options, such as timeZone, `formatTime` acts as `formatDate`. This implies that the following (in Europe at 9:27 the 17. August 2016, CET):

```xml
<FormatTime value={someDate} />
<FormatTime value={someDate} timeZone='America/Los_Angeles' />
```

renders

```xml
<span>9:27 AM</span>
<span>8/17/2016</span>
```

which makes `FormatTime` useless.

This pull request supplies default values `{hour: 'numeric', minute: 'numeric'}` which will make the above example render

```xml
<span>9:27 AM</span>
<span>12:27 AM</span>
```

### DIFF
--- a/src/format.js
+++ b/src/format.js
@@ -85,13 +85,9 @@ export function formatTime(config, state, value, options = {}) {
     let defaults        = format && getNamedFormat(formats, 'time', format);
     let filteredOptions = filterProps(options, DATE_TIME_FORMAT_OPTIONS, defaults);
 
-    // When no formatting options have been specified, default to outputting a
-    // time; e.g.: "9:42 AM".
-    if (Object.keys(filteredOptions).length === 0) {
-        filteredOptions = {
-            hour  : 'numeric',
-            minute: 'numeric',
-        };
+    if (!filteredOptions.hour && !filteredOptions.minute && !filteredOptions.second) {
+        // Add default formatting options if hour, minute, or second isn't defined.
+        filteredOptions = { ...filteredOptions, hour: 'numeric', minute: 'numeric'};
     }
 
     try {
--- a/test/unit/format.js
+++ b/test/unit/format.js
@@ -287,6 +287,38 @@ describe('format API', () => {
                     `[React Intl] No time format named: ${format}`
                 );
             });
+
+            it('should set default values', () => {
+                const date = new Date();
+                const {locale} = config;
+                const day = 'numeric';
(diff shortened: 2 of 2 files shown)


yahoocla:

yahoocla:
CLA is valid!

ericf:
I think this should be fine, while not theoretically fully backwards compatible, it seems likely people won't hit any unexpected issue with this.

src/format.js
```diff
@@ -85,13 +85,9 @@ export function formatTime(config, state, value, options = {}) {
...
     let filteredOptions = filterProps(options, DATE_TIME_FORMAT_OPTIONS, defaults);
 
-    // When no formatting options have been specified, default to outputting a
-    // time; e.g.: "9:42 AM".
-    if (Object.keys(filteredOptions).length === 0) {
-        filteredOptions = {
-            hour  : 'numeric',
-            minute: 'numeric',
-        };
+    if (!filteredOptions.hour && !filteredOptions.minute) {
```
ericf: `second` should also be considered. While not very useful, browsers support the following:
```js
let df = new Intl.DateTimeFormat('en', {second: 'numeric'});
console.log(df.format(Date.now())); // => "30"
```

src/format.js
```diff
@@ -85,13 +85,9 @@ export function formatTime(config, state, value, options = {}) {
...
+        // Add default formatting options if hour or minute isn't defined.
+        filteredOptions = Object.assign({hour: 'numeric', minute: 'numeric'}, filteredOptions);
```
ericf: Use Object rest/spread syntax here `...` instead of depending on `Object.assign` being available in the runtime. The Babel transform will automatically use `Object.assign` of it exists, or an shim helper.

test/unit/format.js
```diff
@@ -287,6 +287,14 @@ describe('format API', () => {
                     `[React Intl] No time format named: ${format}`
                 );
             });
+
+            it('should set default values', () => {
```
ericf: I think it's worth a test to make sure if `hour`, `minute`, or `second` are specified, then the defaults are **not** applied.

ericf:
@budde377 thanks! Merging this now, and release over the weekend or Monday.

ericf:
I released `v2.1.4` with this change.

PR #602, opened 2016-08-19
Fix unit tests for React 15.3.1 change

ericf:
React changed the `React.Children.only()` error message: https://github.com/facebook/react/pull/7514

This updates the unit tests to not rely on testing for the exact message.


yahoocla:
CLA is valid!

PR #606, opened 2016-08-22
Changed readme and hello-world example to use FormattedMessage to eliminate string fragments

yangsu:


yahoocla:
CLA is valid!

ericf:
@yangsu instead of using `<FormattedHTMLMessage>`, this example should use `<FormattedMessage>`: https://github.com/yahoo/react-intl/wiki/Components#formattedmessage

yangsu:
The string contains `<b>` though, that's why i used HTMLMessage

ericf:
The docs I linked to show how to handle that with `<FormattedMessage>`.

ericf:
Thanks!

Issue #513, opened 2016-06-30
Rich text formatting and translations


geekyme:
## Issue

Say i have a string: 

```
To buy a shoe, 
<a class="external_link" target="_blank" href="https://www.shoe.com/">visit our website</a>
and
<strong class="important">eat a shoe</strong>
```

According to https://github.com/yahoo/react-intl/wiki/Components#formattedmessage, I need to format my text this way: 

```
<FormattedMessage
    defaultMessage='To buy a shoe, { link } and { cta }'
    values={{
        link: <a class="external_link" target="_blank" href="https://www.shoe.com/">visit our website</a>,
        cta: <strong class="important">eat a shoe</strong>
    }}
/>
```

When doing string extraction with https://github.com/yahoo/babel-plugin-react-intl, I would get:

 `To buy a shoe, { link } and { cta }` to be translated. 

This isn't good because I lose the context and texts from the rich content for `link` and `cta`. 
## Suggestion

I think we should have some special delimiters to surround rich text content so translators can still translate the text as a whole:

`To buy a shoe, @visit our website@ and @eat a shoe@`

Thoughts?

ericf:
@geekyme can you explain more how this would work?

geekyme:
What I have in mind is `FormattedMessage` should look at the string presented in `defaultMessage`, and if it finds an opening and closing `@`, it takes the contents between them and look up a list of rich markups and interpolate the text inside the markup. 

Eg. `visit our website` gets interpolated between `<a class="external_link" target="_blank" href="https://www.shoe.com/"></a>`. This can be done using `React.cloneElement` then providing the text as children. Same thing for `eat a shoe`

```
<FormattedMessage
    defaultMessage='To buy a shoe, @visit our website@ and @eat a shoe@'
   interpolate={[
       <a class="external_link" target="_blank" href="https://www.shoe.com/"></a>,
       <strong class="important">eat a shoe</strong>
   ]}
/>
```

ericf:
This sounds related to: https://github.com/yahoo/react-intl/issues/455#issuecomment-218282929

Trying to make your example work can quickly break down, imagine if someone had:
```html
<strong>eat a <em>shoe</em></strong>
```

azmenak:
Came across a similar problem today, maybe there's a better way to do this that I'm not seeing at the moment.

If I want to "format" links in a message, the only option I see right now is passing an additional `<FormattedMessage />` in the `values` of the parent. IE.

```jsx
[...]

PR #612, opened 2016-08-23
Spike for shorthand `message`

ericf:
This is a quick spike to show how a new `message` export would create a Message Descriptor and capture the `values` for formatting. This approach would be the same weather you apply the descriptor to `<FormattedMessage>` or `formatMessage()`.

See: #609


yahoocla:
CLA is valid!

ericf:
Added message extraction support for this in https://github.com/yahoo/babel-plugin-react-intl/pull/70

oscar-b:
This looks really nice, any update on its status @ericf ?

ericf:
@oscar-b I'm working on https://github.com/yahoo/react-intl/issues/513#issuecomment-252114827 which sort of supersedes this work. There might be something that's still relevant here which can be applied as well.

Billy-:
@ericf is this package and https://github.com/yahoo/babel-plugin-react-intl still being worked on? Seems like it's been abandoned...?

esquevin:
The package doesn't seem to be updated. If it's possible for you I'd advise you to try to use [react-i18nnext](https://github.com/i18next/react-i18next) instead

PR #673, opened 2016-09-28
Port react-router examples to use Create React App

iswanj:
@ericf Here I added CRA for react-router example. Also, Other 3 example ( async, nested, translation ) are little tricky to add CRA, because of the structure of examples. 
So I'll add an another PR for above examples later.
.


yahoocla:
CLA is valid!

ericf:
@iswanj I wonder if it's worth using `react-router@4` for this? https://react-router.now.sh

iswanj:
@ericf :) Ok I'll add react-router@4

iswanj:
@ericf Added react-router v4

ericf:
@iswanj Thanks! 🎉

PR #685, opened 2016-10-11
Update eslint and add rules for eol and jsx-indent

zouxuoz:
I'm updated eslint and eslint-plugin-react, added rules for check eol and jsx-index and added command for fix eslint errors.


yahoocla:
CLA is valid!

zouxuoz:
@ericf build failed because eslint 3.x hasn't support NodeJS 0.12 because [LTS](https://github.com/nodejs/LTS) will soon end. I can downgrade eslint back or you can remove NodeJS 0.12 from supported versions. What are you thinking about it?

ericf:
@zouxuoz I want to keep testing on Node 0.12 since I still want to support it and fix any bugs that break in Node 0.12.

zouxuoz:
@ericf I'm downgrade eslint to latest 2.x version and fix config for that.

ericf:
Thanks!

Issue #687, opened 2016-10-14
Minimize bundled code size


mbrevda:
It seems that all modules are bundled into a single dist file, leading to ~31k of imports. This is true even if only some of the packages functionality is used. Other packages, such as [react-router](https://github.com/ReactTraining/react-router/blob/v4/website/examples/Basic.js#L2-L4) and [react-virtualized](https://github.com/bvaughn/react-virtualized/issues/427#issuecomment-252367465) have taken to providing direct module imports where the indevidual modules are built and can be included without importing the entire package.

In the name of saving bytes, I propose breaking down react-intl's components, if possible, to separate modules. For the sake of backwards compatibility, the current single file import would stay, with sinlge modules files complimenting it.

ericf:
Isn't this moot with Rollup and Webpack 2 once the issues is resolved in Rollup between `"module"` and `"browser"` in `package.json`?

mbrevda:
The utopia of perfect tree shaking seems to be far off even for those that **are** using rollup/webpack (and not everybody is...). [See here](https://github.com/ReactTraining/react-router/pull/3410) for more discussion. In the meantime, we have direct imports.

mbrevda:
Is lib size not a priority for this project?

ericf:
This isn't actionable without a major version bump. I'd rather setup specific issues related to actions that can happen for the current major version. I have some ideas on how to do that.

mbrevda:
Great to hear, looking forward to seeing what you come up with!

jacobgreen25:
Would greatly appreciate this @ericf. Thanks @mbrevda for bringing this up.

Issue #549, opened 2016-07-26
Please export all library propTypes


steida:
https://github.com/yahoo/react-intl/blob/master/src/types.js
For example, I need messageDescriptorPropTypes.

Thank you.

ericf:
I curious about the use case. Right now `intlShape` is the only public propTypes that are exposed. In the particular case of `messageDescriptorPropTypes`, it's fairly simple to redefine.

steida:
That's what I did, but I would rather avoid reimplementation. You can close this issue, it's not critical of course :-)

ericf:
Well I am interested in how you were using the prop types definition. Of all of the possible ones to expose from `types.js` this would be at the top of the list.

steida:
Check Alert component, it takes message as a prop. https://github.com/este/este/blob/master/src/native/app/components/Alert.react.js#L39

paulyoung:
I also would like to use `messageDescriptorPropTypes`.

steida:
PropTyped are dead. Or at least, I don't need them.

fikriauliya:
@steida Why is it dead?

steida:
Because flowtype.

kusmierz:
see flowtype/flow-typed#1280

PR #696, opened 2016-10-19
Remove propTypes from production builds

mbrevda:
Motivation
===
Removed propTypes from production builds, as they are unused by React in production and just take up extra bytes. This can reduce the minimized dist build by as much as **~3%!**:

```
// before
48170 Oct 19 11:21 react-intl.min.js

// after
46836 Oct 19 11:24 react-intl.min.js
```

What changed
===
For `lib/` builds, propTypes are wrapped in a conditional statement (`if(process.env.NODE_ENV !== 'production'){}`). This keeps the proptypes in place during development so that dev have access to them, but removes them when built with a packager (browserify/webpack/rollup/etc) that replaces `process.env.NODE_ENV` with the actual value.

For `dist/` build, the process is similar, but when `isProduction == true`, the propTypes will be completely removed. It was necessary to move the `replace()` plugin to the `isProduction == true` build so that propTypes aren't completely removed for non production build.

Finally, due to the non-standard way of how propTypes are defined in this project, the definitions themselves were manually wrapped where possible (see below). This, again, keeps them in development build but removes them for production builds.

TODO:
===
It seems some proptypes are used by the code during runtime for non-propType related activity. Can these be updated/removed/otherwise minified for even more saving?

Disclaimer
===
I hope I didn't misunderstand how the build process works!

Prior art:
===
 * react-virtualized: https://github.com/bvaughn/react-virtualized/pull/438 & https://github.com/bvaughn/react-virtualized/pull/440
 * react-router: https://github.com/ReactTraining/react-router/issues/4046
 * react: https://github.com/facebook/react/pull/7651, https://github.com/facebook/react/pull/2948, amongst others
 * the plugin: https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types

Related:
===
This is a first step towards https://github.com/yahoo/react-intl/issues/687

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -7,6 +7,10 @@
 import {PropTypes} from 'react';
 
 const {bool, number, string, func, object, oneOf, shape} = PropTypes;
+const localeMatcher = oneOf(['best fit', 'lookup']);
+const narrowShortLong = oneOf(['narrow', 'short', 'long']);
+const numeric2digit = oneOf(['numeric', '2-digit']);
+const funcReq = func.isRequired;
 
 export const intlConfigPropTypes = {
     locale  : string,
@@ -18,20 +22,20 @@ export const intlConfigPropTypes = {
 };
 
 export const intlFormatPropTypes = {
-    formatDate       : func.isRequired,
-    formatTime       : func.isRequired,
-    formatRelative   : func.isRequired,
-    formatNumber     : func.isRequired,
-    formatPlural     : func.isRequired,
-    formatMessage    : func.isRequired,
-    formatHTMLMessage: func.isRequired,
+    formatDate       : funcReq,
+    formatTime       : funcReq,
+    formatRelative   : funcReq,
+    formatNumber     : funcReq,
+    formatPlural     : funcReq,
+    formatMessage    : funcReq,
+    formatHTMLMessage: funcReq,
 };
 
 export const intlShape = shape({
     ...intlConfigPropTypes,
     ...intlFormatPropTypes,
     formatters: object,
-    now: func.isRequired,
+    now: funcReq,
 });
 
 export const messageDescriptorPropTypes = {
@@ -41,25 +45,25 @@ export const messageDescriptorPropTypes = {
 };
 
 export const dateTimeFormatPropTypes = {
-    localeMatcher: oneOf(['best fit', 'lookup']),
+    localeMatcher,
     formatMatcher: oneOf(['basic', 'best fit']),
 
     timeZone: string,
     hour12  : bool,
 
-    weekday     : oneOf(['narrow', 'short', 'long']),
-    era         : oneOf(['narrow', 'short', 'long']),
-    year        : oneOf(['numeric', '2-digit']),
+    weekday     : narrowShortLong,
+    era         : narrowShortLong,
+    year        : numeric2digit,
     month       : oneOf(['numeric', '2-digit', 'narrow', 'short', 'long']),
-    day         : oneOf(['numeric', '2-digit']),
-    hour        : oneOf(['numeric', '2-digit']),
-    minute      : oneOf(['numeric', '2-digit']),
-    second      : oneOf(['numeric', '2-digit']),
+    day         : numeric2digit,
+    hour        : numeric2digit,
+    minute      : numeric2digit,
+    second      : numeric2digit,
     timeZoneName: oneOf(['short', 'long']),
 };
 
 export const numberFormatPropTypes = {
-    localeMatcher: oneOf(['best fit', 'lookup']),
+    localeMatcher,
 
     style          : oneOf(['decimal', 'currency', 'percent']),
     currency       : string,
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "babel-plugin-transform-es3-member-expression-literals": "^6.3.13",
     "babel-plugin-transform-es3-property-literals": "^6.3.13",
     "babel-plugin-transform-object-rest-spread": "^6.1.18",
+    "babel-plugin-transform-react-remove-prop-types": "^0.2.10",
     "babel-polyfill": "^6.3.14",
     "babel-preset-es2015": "^6.1.18",
     "babel-preset-es2015-rollup": "^1.1.1",
--- a/src/.babelrc
+++ b/src/.babelrc
@@ -7,7 +7,8 @@
         "transform-object-rest-spread",
         "transform-class-properties",
         "transform-es3-member-expression-literals",
-        "transform-es3-property-literals"
+        "transform-es3-property-literals",
+        ["transform-react-remove-prop-types", {"mode": "wrap"}]
     ],
     "env": {
         "test": {


yahoocla:

yahoocla:
CLA is valid!

scripts/build-dist.js
```diff
@@ -57,13 +59,13 @@ let plugins = [
...
     }),
-    replace({
-        'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
-    }),
 ];
 
 if (isProduction) {
     plugins.push(
+        replace({
+            'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV),
```
ericf: What did you move this into the the production build only?

scripts/build-dist.js
```diff
@@ -45,6 +45,8 @@ babelConfig.babelrc = false;
 babelConfig.presets = babelConfig.presets.map((preset) => {
     return preset === 'es2015' ? 'es2015-rollup' : preset;
 });
+babelConfig.plugins = babelConfig.plugins || [];
+babelConfig.plugins.push(['transform-react-remove-prop-types', {'mode': 'wrap'}]);
```
ericf: I think a better approach would be to add this to the `.babelrc` when `"env"` is `"production"`.

ericf: Why `mode: 'wrap'`, instead of the default to remove?

mbrevda:
> Yes, I'd like the leave the types.js file alone for this reason.

Leaving types.js alone mostly inverses the (already tiny) gains we are seeking here. I suggest modifying it smartly and that which can be eliminates should be, as I have done in the PR.

> I'd like to know what the gzip savings is of this change

I don't know. Even gzipped code has a price during unzipping and parsing. While these changes probably won't add much weight in either direction, every byte counts.

> Does this Babel plugin support removing static proptType = {}; from the components? 

yes, it does. However some have argued that leaving them in during development is valuable to developers (otherwise, why include them at all?). The middle of the road seems to be: ship them so that dev's can leverage them, but add a simple way to strip them out for production (see links above for more).

> allow Rollup to do its tree-shaking magic

As I mentioned in #687, tree-shaking is more smoke and mirrors then magic. While it mostly delivers for library authors, it really disappoints when working on a real life project. Currently the react project advocates the `process.env.NODE_ENV !== 'production'` method (and makes heavy use of it, making that the lowest common denominator for ensuring the minimal amount of code is being shipped.

scripts/build-dist.js
mbrevda: see https://github.com/yahoo/react-intl/pull/696#issuecomment-254942556, 3rd quote

scripts/build-dist.js
mbrevda: We could move it to `.babelrc`, but it doesn't seem like your currently setup for env builds -  and thats beyond the scope of this PR.

ericf: The scripts are doing builds for both development and production. I'm suggesting that you do what's in the README for this Babel plugin: https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types#via-babelrc-recommended

ericf:
I quickly tried what I suggested, which required no code changes, just `package.json` and `.babelrc` changes and got these results:

**Before:**
```
48170 Oct 19 18:45 react-intl.min.js
13363 Oct 19 18:45 react-intl.min.js.gz
```

**After:**
```
47025 Oct 19 18:41 react-intl.min.js
13206 Oct 19 18:41 react-intl.min.js.gz
```

So **2.4%** savings on `.min.js` and **1.2%** savings on `.min.js.gz`.

As for reducing `src/types.js` in production, I have other requests asking to _expose_ that entire module for them to use: https://github.com/yahoo/react-intl/issues/549

Here's my diff:
```diff
diff --git a/package.json b/package.json
index 7b044b4..8721ab7 100644
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "babel-plugin-transform-es3-member-expression-literals": "^6.3.13",
     "babel-plugin-transform-es3-property-literals": "^6.3.13",
     "babel-plugin-transform-object-rest-spread": "^6.1.18",
+    "babel-plugin-transform-react-remove-prop-types": "^0.2.10",
     "babel-polyfill": "^6.3.14",
     "babel-preset-es2015": "^6.1.18",
     "babel-preset-es2015-rollup": "^1.1.1",
diff --git a/src/.babelrc b/src/.babelrc
index b80ddf1..8412546 100644
--- a/src/.babelrc
+++ b/src/.babelrc
@@ -14,6 +14,11 @@
             "plugins": [
                 "istanbul"
             ]
+        },
+        "production": {
+            "plugins": [
+                "transform-react-remove-prop-types"
+            ]
         }
     }
 }
```

ericf:
btw, Rollup is smart enough to tree-shake the defs in `src/types.js` that aren't being used in the prod code; e.g. `messageDescriptorPropTypes`.

ericf:
You could also define a few of the repetitive types so they don't need to be duplicated in `src/types.js`:
```diff
diff --git a/src/types.js b/src/types.js
index ffad182..12d3a1f 100644
--- a/src/types.js
+++ b/src/types.js
@@ -40,26 +40,30 @@ export const messageDescriptorPropTypes = {
     defaultMessage: string,
 };
 
+const localeMatcher = oneOf(['best fit', 'lookup']);
+const narrowShortLong = oneOf(['narrow', 'short', 'long']);
+const numeric2digit = oneOf(['numeric', '2-digit']);
+
 export const dateTimeFormatPropTypes = {
-    localeMatcher: oneOf(['best fit', 'lookup']),
+    localeMatcher: localeMatcher,
     formatMatcher: oneOf(['basic', 'best fit']),
 
     timeZone: string,
     hour12  : bool,
 
-    weekday     : oneOf(['narrow', 'short', 'long']),
-    era         : oneOf(['narrow', 'short', 'long']),
-    year        : oneOf(['numeric', '2-digit']),
+    weekday     : narrowShortLong,
+    era         : narrowShortLong,
+    year        : numeric2digit,
     month       : oneOf(['numeric', '2-digit', 'narrow', 'short', 'long']),
-    day         : oneOf(['numeric', '2-digit']),
-    hour        : oneOf(['numeric', '2-digit']),
-    minute      : oneOf(['numeric', '2-digit']),
-    second      : oneOf(['numeric', '2-digit']),
+    day         : numeric2digit,
+    hour        : numeric2digit,
+    minute      : numeric2digit,
+    second      : numeric2digit,
     timeZoneName: oneOf(['short', 'long']),
 };
 
 export const numberFormatPropTypes = {
-    localeMatcher: oneOf(['best fit', 'lookup']),
+    localeMatcher: localeMatcher,
 
     style          : oneOf(['decimal', 'currency', 'percent']),
     currency       : string,
```

mbrevda:
I have made some of the suggested changes. Were currently looking at:
```
46732 Oct 20 10:40 react-intl.min.js
13123 Oct 20 10:40 react-intl.min.js.gz
```

What other changes do you feel are necessary or can be usefull?

As file compression works better on larger datasets, the gz size isn't fully indicative (as I'm sure you know).

src/types.js
```diff
@@ -7,59 +7,62 @@
...
+const localeMatcher = oneOf(['best fit', 'lookup']);
+const narrowShortLong = oneOf(['narrow', 'short', 'long']);
+const numeric2digit = oneOf(['numeric', '2-digit']);
+const funcReq = func.isRequired;
 
 export const intlConfigPropTypes = {
     locale  : string,
     formats : object,
     messages: object,
-
```
ericf: Please add back the line break here.

src/types.js
```diff
@@ -7,59 +7,62 @@
...
 
-export const intlFormatPropTypes = {
-    formatDate       : func.isRequired,
-    formatTime       : func.isRequired,
-    formatRelative   : func.isRequired,
-    formatNumber     : func.isRequired,
-    formatPlural     : func.isRequired,
-    formatMessage    : func.isRequired,
-    formatHTMLMessage: func.isRequired,
+export const intlFormatPropTypes =  {
```
ericf: Remove the added space.

src/types.js
```diff
@@ -7,59 +7,62 @@
...
+    formatTime       : funcReq,
+    formatRelative   : funcReq,
+    formatNumber     : funcReq,
+    formatPlural     : funcReq,
+    formatMessage    : funcReq,
+    formatHTMLMessage: funcReq,
 };
 
-export const intlShape = shape({
+export const intlShape = process.env.NODE_ENV !== 'production' ? shape({
```
ericf: Please remove these ternaries for the reasons stated in the other comments.

src/types.js
```diff
@@ -7,59 +7,62 @@
...
     ...intlConfigPropTypes,
     ...intlFormatPropTypes,
     formatters: object,
-    now: func.isRequired,
-});
+    now: funcReq,
+}) : ()=>{};
 
-export const messageDescriptorPropTypes = {
+export const messageDescriptorPropTypes = process.env.NODE_ENV !== 'production' ? {
```
ericf: Please remove these ternaries for the reasons stated in the other comments.

src/types.js
```diff
@@ -73,11 +76,11 @@ export const numberFormatPropTypes = {
     maximumSignificantDigits: number,
 };
 
-export const relativeFormatPropTypes = {
+export const relativeFormatPropTypes = process.env.NODE_ENV !== 'production' ? {
```
ericf: Please remove these ternaries for the reasons stated in the other comments.

src/types.js
```diff
@@ -73,11 +76,11 @@ export const numberFormatPropTypes = {
...
     style: oneOf(['best fit', 'numeric']),
     units: oneOf(['second', 'minute', 'hour', 'day', 'month', 'year']),
-};
+} : {};
 
-export const pluralFormatPropTypes = {
+export const pluralFormatPropTypes = process.env.NODE_ENV !== 'production' ? {
```
ericf: Please remove these ternaries for the reasons stated in the other comments.

mbrevda:
> I was literally stating a fact

Got it!

Did I miss anything else?

scripts/build-dist.js
ericf: @mbrevda can you revert this change to `scripts/built-dist.js` too?

ericf:
> Did I miss anything else?

@mbrevda can you revert the changes to `scripts/build-dist.js` as well? https://github.com/yahoo/react-intl/pull/696#discussion_r84331248

mbrevda:
I'm pretty sure that would also prevent propTypes from being utilized by
devs during development. Are you certain you want me to revert the change?

ericf:
> I'm pretty sure that would also prevent propTypes from being utilized by
> devs during development. Are you certain you want me to revert the change?

They are there in development. And you already reverted the change to `scripts/build-dist.js` before I merged this.

mbrevda:
I'm pretty sure that would also prevent propTypes from being utilized by
devs during development - the only time propTypes could be used. Are you
certain you want that reverted?

On Thu, Oct 20, 2016, 9:30 PM Eric Ferraiuolo <notifications@github.com>
wrote:

Did I miss anything else?

@mbrevda <https://github.com/mbrevda> can you revert the changes to
scripts/build-dist.js as well? #696 (comment)
<https://github.com/yahoo/react-intl/pull/696#discussion_r84331248>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/yahoo/react-intl/pull/696#issuecomment-255189182>, or mute
the thread

mbrevda:
Hmm - seems github just posted emails sent a week ago. Ignore them!

ericf:
Ha, lol, I was wondering what was going on 😄

PR #701, opened 2016-10-21
Remove propTypes from production builds

ericf:
DNS issues today caused GitHub <--> Travis web hook issues, so this is a dup of #696 to get CI to run.


yahoocla:
CLA is valid!

PR #703, opened 2016-10-23
Use defaultFormats in formatNumber, formatDate, formatTime and formatRelative

### LINKED ISSUE #702: Ability to explicitly pass timeZone to IntlProvider
I'd like to be able to set default formats for the entire application which are used by each of the different format functions. Currently, when passing defaultFormats to IntlProvider, these formats are only used by the formatMessage function.

For example, if I want all times displayed in a specific time zone (namely Africa/Johannesburg), it would be useful to pass the following to IntlProvider:

```
    return (
      <IntlProvider
        defaultFormats={{
          time: {
            timeZone: 'Africa/Johannesburg'
          }
        }}>
...
```

brichardssa:
resolve #702 

There may be times when default format options for numbers, dates and times need to be set explicitly by passing them to IntlProvider via props.

Added default-format example to demonstrate new functionality

### DIFF
--- a/examples/default-formats/.gitignore
+++ b/examples/default-formats/.gitignore
@@ -1,0 +1,15 @@
+# See http://help.github.com/ignore-files/ for more about ignoring files.
+
+# dependencies
+node_modules
+
+# testing
+coverage
+
+# production
+build
+
+# misc
+.DS_Store
+.env
+npm-debug.log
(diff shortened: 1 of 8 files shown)


yahoocla:
CLA is valid!

StorytellerCZ:
Superseded by #893

redonkulus:
Closed by https://github.com/yahoo/react-intl/pull/893

PR #739, opened 2016-11-16
Adaptations for React Native

mschipperheyn:
WIP


yahoocla:
CLA is valid!

ericf:
It should be `textComponent` instead of `textElement` since you're passing around a component name or function.

I'm wondering if it makes more sense to limit the option on `<IntlProvider>` to accept something like `isReactNative` and keep the `textComponent` a private thing out of the public API. This way in the future when there's a better way to handle this, `isReactNative` can become a noop, and with React Fiber we'll be able to simply return a string element without a wrapper and not have to support `textComponent` as a public thing.

mschipperheyn:
The isReactNative approach is imho not a good approach. Unlike with the DOM, RN is not good at passing styles down the tree, so if you want to guarantee that the rendered text looks the way you want, you will need to pass down a custom Text element that contains the styling you need.

ericf:
I thought nested `<Text>` elements do propagate styles? The usage pattern would be:

```js
<Text style={style.text}>
  <FormattedNumber value={1000}/>
</Text>
```
Which would create:
```js
<Text style={styles.text}>
  <Text>1,000</Text>
</Text>
```

mschipperheyn:
Yes, fair enough, that should work. Personally, I would prefer passing down my own component one time instead of adding additional surrounding tags.

ericf:
> Text supports nesting, styling, and touch handling.
> 
> In the following example, the nested title and body text will inherit the fontFamily from styles.baseText, but the title provides its own additional styles. The title and body will stack on top of each other on account of the literal newlines:

https://facebook.github.io/react-native/docs/text.html

ericf:
> Yes, fair enough, that should work. Personally, I would prefer passing down my own component one time instead of adding additional surrounding tags.

But you'd need to create your own `<FancyText>` component for the rest of your app anyways, so what's different of wrapping `<FancyText>` around `"foo bar"` or around a React Intl component?

I'm just trying to make React Intl work in React Native without changing the programming model and maximizing the future backwards compatibility 😄

mschipperheyn:
Ah ok, yeah, but this idea of propagating styles down the structure only works if there is an uninterrupted structure of Texts. So, my understanding of what you're proposing is that I would have to surround every FormattedMessage with a Text style=[...] 

With what I'm suggesting, you only pass it to IntlProvider once and then just use FormattedMessage without surrounding tags.

mschipperheyn:
This is what my AText component looks like

		'use strict';

		import React from 'react';
		import { Text, StyleSheet } from 'react-native';
		import { styleVars } from '../styles/styles';

		export default class AText extends React.PureComponent{

		  render(){
			return <Text {...this.props} style={[localStyle.text, this.props.style]}>{this.props.children}</Text>
		  }
		}

		const localStyle = StyleSheet.create({
		  text:{
			fontFamily:styleVars.fontFamily,
			fontSize:styleVars.fontSize
		  }
		});

wangsha:
+1

huhuanming:
+1

huhuanming:
0 w 0 I love it!!!

huhuanming:
+2

huhuanming:
+3

mschipperheyn:
https://github.com/yahoo/react-intl/issues/600#issuecomment-267050947

src/components/date.js
```diff
@@ -32,7 +32,7 @@ export default class FormattedDate extends Component {
     }
 
     render() {
-        const {formatDate}      = this.context.intl;
+        const {formatDate, textComponent} = this.context.intl;
```
ericf: I would prefer if this were:
```js
const {formatDate, textComponent: Text} = this.context.intl;
```
…and then keep the JSX at the usage site.

src/components/date.js
```diff
@@ -41,6 +41,6 @@ export default class FormattedDate extends Component {
             return children(formattedDate);
         }
 
-        return <span>{formattedDate}</span>;
+        return React.createElement(textComponent, undefined, formattedDate);
```
ericf: Based on the previous comment, you could keep the JSX and it would be:
```js
<Text>{formattedDate}</Text>
```

src/components/number.js
```diff
@@ -41,6 +41,6 @@ export default class FormattedNumber extends Component {
             return children(formattedNumber);
         }
 
-        return <span>{formattedNumber}</span>;
+        return React.createElement(textComponent, undefined, formattedNumber);
```
ericf: Same comments as `<FormattedDate>`.

src/components/plural.js
```diff
@@ -53,6 +53,6 @@ export default class FormattedPlural extends Component {
             return children(formattedPlural);
         }
 
-        return <span>{formattedPlural}</span>;
+        return React.createElement(textComponent, undefined, formattedPlural);
```
ericf: Same comments as `<FormattedDate>`.

src/components/provider.js
```diff
@@ -23,6 +23,7 @@ const intlFormatPropNames = Object.keys(intlFormatPropTypes);
 const defaultProps = {
     formats : {},
     messages: {},
+    textComponent: 'span',
```
ericf: This seems good. And if later, in another PR, if there's a good idiomatic way to detect React Native vs React DOM, then this can be updated to be dynamic.

src/components/relative.js
```diff
@@ -158,6 +158,6 @@ export default class FormattedRelative extends Component {
             return children(formattedRelative);
         }
 
-        return <span>{formattedRelative}</span>;
+        return React.createElement(textComponent, undefined, formattedRelative);
```
ericf: Same comments as `<FormattedDate>`.

src/components/time.js
```diff
@@ -41,6 +41,6 @@ export default class FormattedTime extends Component {
             return children(formattedTime);
         }
 
-        return <span>{formattedTime}</span>;
+        return React.createElement(textComponent, undefined, formattedTime);
```
ericf: Same comments as `<FormattedDate>`.

src/types.js
```diff
@@ -16,6 +16,7 @@ export const intlConfigPropTypes = {
     locale  : string,
     formats : object,
     messages: object,
+    textComponent: node,
```
ericf: Perfect! 👍

huhuanming:
+3

mschipperheyn:
Ok, I wil process these changes

huhuanming:
awesome !!!!!

src/components/date.js
mschipperheyn: I guess I don't understand this. if the textComponent is of type Text, it would become RN specific. Did I misunderstand something? Shouldn't it be of type `node`?

mschipperheyn: Ah ok. I thought we were talking about a type here. but you are talking about creating an alias

src/components/provider.js
mschipperheyn: Yeah, although I think given our earlier discussion on the propagation of css properties, it's good to be careful with automatic detection.

mschipperheyn:
Ok, applied the changes. For consistency, also created the 'Text' alias in areas where createElement was used anyways (e.g. message.js)

src/components/date.js
ericf: Yeah, alias.

ericf:
@mschipperheyn thanks! I'll work on releasing this today.

oun:
Hi romanenko. I got the same warning. Did you find any solutions?

romanenko:
@oun I've upgraded react-intl to `v2.2.3` and the problem is gone.

PR #786, opened 2016-12-24
Fix failed prop type warnings on React Native

fson:
Using `react-intl@2.2.0` with React Native as documented (with `<IntlProvider textComponent={Text} ...>`) shows a prop type warning: "Warning: Failed prop type: Invalid prop `textComponent` supplied to `IntlProvider`"

The reason is that `Text` is not `node` but a component class. The correct `propType` for `textComponent` is "React component/function/class", but since this prop type doesn't exist, loosen it to `any`.

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -6,7 +6,7 @@
 
 import {PropTypes} from 'react';
 
-const {bool, number, string, func, object, oneOf, shape, node} = PropTypes;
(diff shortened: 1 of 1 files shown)


yahoocla:

fson:
I've signed the CLA.

PR #795, opened 2016-12-29
Fix key prop warning

### LINKED ISSUE #787: Regression: missing "key" prop warning
Hi, guys!

I've updated to 2.2.0 and noticed that when I provide a message as a value to another message, React started throwing warning regarding missing key prop.

Here are a couple of jsbins to show you what I mean, have a look in the console

2.1.5: http://jsbin.com/jumogusomo/2/edit?html,js,output
2.2.0: http://jsbin.com/jumogusomo/3/edit?html,js,output

Can you have a look, please?

ericf:
This reverts back to using `createElement` instead of JSX inside the `render()` method of `<FormattedMessage>` to remove the missing `key` prop warning that React is throwing in `react-intl@2.2.0`.

Fixes #787

### DIFF
--- a/src/components/message.js
+++ b/src/components/message.js
@@ -4,7 +4,7 @@
  * See the accompanying LICENSE file for terms.
  */
 
-import React, {Component, PropTypes, isValidElement} from 'react';
+import {Component, PropTypes, createElement, isValidElement} from 'react';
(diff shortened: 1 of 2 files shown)


yahoocla:
CLA is valid!

PR #819, opened 2017-01-17
Prevent <FormattedRelative> from updating with bad `value`

ericf:
When an invalid `value` is provided to `<FormattedRelative>` this change makes sure its update scheduler doesn't get stuck in a tight loop.


yahoocla:
CLA is valid!

Nagarkarricha:
👍

kfay:
👍

PR #839, opened 2017-02-05
Made line-ending a warning.  It's still an annoyance for windows user…

OracPrime:
…s, but it doesn't stop scripts running.

Added .cmd extents to allow examples.js to run on windows.
Made the hello world example actually use a non-standard locale so that you can see that something is happening.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,7 +22,7 @@
         "eqeqeq": 2,
         "eol-last": 2,
         "indent": [2, 4],
-        "linebreak-style": [2, "unix"],
+        "linebreak-style": [1, "unix"],
         "prefer-arrow-callback": 2,
         "quotes": [2, "single", "avoid-escape"],
         "semi": [2, "always"],
--- a/examples/hello-world/src/App.js
+++ b/examples/hello-world/src/App.js
@@ -1,11 +1,11 @@
-import React, {Component} from 'react';
-import {FormattedMessage,FormattedRelative} from 'react-intl';
+import React, { Component } from 'react';
+import { FormattedMessage, FormattedRelative } from 'react-intl';
 
 class App extends Component {
     constructor(props) {
         super(props);
         this.state = {
-            name       : 'Eric',
+            name: 'Eric',
             unreadCount: 1000,
         };
     }
@@ -15,17 +15,17 @@ class App extends Component {
 
         return (
             <div>
-            <p>
-                <FormattedMessage
-                    id="welcome"
-                    defaultMessage={`Hello {name}, you have {unreadCount, number} {unreadCount, plural,
+                <p>
+                    <FormattedMessage
+                        id="welcome"
+                        defaultMessage={`Hello {name}, you have {unreadCount, number} {unreadCount, plural,
                       one {message}
                       other {messages}
                     }`}
-                    values={{name: <b>{name}</b>, unreadCount}}
-                />
-            </p>
-            <p> Last update <FormattedRelative value={Date.now()-3600000}/> </p>
+                        values={{ name: <b>{name}</b>, unreadCount }}
+                        />
+                </p>
+                <p> Last update <FormattedRelative value={Date.now() - 3600000} /> </p>
             </div>
         );
     }
--- a/examples/hello-world/src/App.js
+++ b/examples/hello-world/src/App.js
@@ -1,11 +1,11 @@
-import React, { Component } from 'react';
-import { FormattedMessage, FormattedRelative } from 'react-intl';
+import React, {Component} from 'react';
+import {FormattedMessage} from 'react-intl';
 
 class App extends Component {
     constructor(props) {
         super(props);
         this.state = {
-            name: 'Eric',
+            name       : 'Eric',
             unreadCount: 1000,
         };
     }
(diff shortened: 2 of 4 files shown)


yahoocla:

OracPrime:
If you mean the changes to examples/hello-world/src/index.js and App.js , I made those because without them it was not obvious to me that the internationalisation was working on Windows - ie. no translation was being performed.  So if we modify the scope slightly to "make the examples *obviously* work on Windows" then I think it's in scope.  I guess if you don't want to "prefer" French in any way, you could change index.js to render
<div>
    <IntlProvider locale="en">
        <App />
    </IntlProvider>
    <IntlProvider locale="fr">
        <App />
    </IntlProvider>
</div>

The examples/hello-world changes aren't necessary, but I felt they would add to the understanding of a new windows developer coming to the project.  I won't be at all upset if you just take the other changes!

.eslintrc
```diff
@@ -22,7 +22,7 @@
         "eqeqeq": 2,
         "eol-last": 2,
         "indent": [2, 4],
-        "linebreak-style": [2, "unix"],
+        "linebreak-style": [1, "unix"],
```
ericf: Re: your comment in the issue about Windows and Git… can you explain a bit more of what's happening now? Is there a way to have a Windows environment that uses unix line endings?

examples/hello-world/src/App.js
```diff
@@ -1,11 +1,11 @@
-import React, {Component} from 'react';
-import {FormattedMessage} from 'react-intl';
+import React, { Component } from 'react';
+import { FormattedMessage, FormattedRelative } from 'react-intl';
```
ericf: Can you revert these formatting changes?

examples/hello-world/src/App.js
```diff
@@ -14,16 +14,19 @@ class App extends Component {
...
                       one {message}
                       other {messages}
                     }`}
-                    values={{name: <b>{name}</b>, unreadCount}}
-                />
-            </p>
+                        values={{ name: <b>{name}</b>, unreadCount }}
+                        />
+                </p>
+                <p> Last update <FormattedRelative value={Date.now() - 3600000} /> </p>
```
ericf: Using `Date.now()` inside `render()` is an anti-pattern that I don't what to encourage in the examples.

examples/hello-world/src/index.js
```diff
@@ -1,10 +1,11 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import {IntlProvider} from 'react-intl';
+import {IntlProvider,addLocaleData} from 'react-intl';
 import App from './App';
-
+import fr from 'react-intl/locale-data/fr';
+addLocaleData(fr);
```
ericf: I want to avoid using more than one language for the hello-world example. React Intl is still useful even if you're only using English.

scripts/examples.js
```diff
@@ -8,8 +8,8 @@ globSync('./examples/*/').forEach((exampleDir) => {
         cwd: exampleDir,
         stdio: 'inherit',
     };
-
-    const result = spawnSync(command, args, opts);
+    let envCommand=command+ (/^win/.test(process.platform) ? '.cmd' : '');
```
ericf: I think this formatting is a bit confusing. Can you switch it to this (spaces around `=` and `+`:
```js
let envCommand = command + (/^win/.test(process.platform) ? '.cmd' : '');
```

.eslintrc
OracPrime: When git pulls from GitHub to Windows, it converts the linebreaks to Windows-style.  When you push back from Windows it converts back.  So (at least in theory) Windows and Linux developers each see their own line endings, but the repository only sees Linux line endings.  So having Linux and the CI builds check line endings is fine, but it breaks the tests when run by a Windows developer.  There are discussions at eslint about having environment-dependent rules, but they haven't come to anything yet.  So if we make it a warning, the CI and Linux developers will see there is a problem (if it gets past the auto-convert from Git) in that they will get any warnings at all.  Windows developers, second rate citizens that we are :), will just have to ignore the multiple-thousand warnings, but at least it won't make pre-commit tests fail.

scripts/examples.js
OracPrime: Sure.

examples/hello-world/src/index.js
OracPrime: Fair enough.  Maybe when someone has time we can add a multi-language demo which demonstrates multiple languages on one page.  Although the irony of wanting Hello World to be in a single language is not lost on me :)

examples/hello-world/src/App.js
OracPrime: Good point.  I'll slap my own wrist for that one.  And revert all the hello-world changes.

.eslintrc
ericf: Okay, cool. So what you have seems like the best solution until ESLint does something about it.

scripts/examples.js
OracPrime: I'm also talking to the cross-env guys about why the nested example fail on Windows.  It's not clear whether cross-env or cross-spawn needs correcting, but the fault lies there, not here.

OracPrime: OK, I think all those changes are in now.

ericf:
@OracPrime thanks! Merged!

PR #843, opened 2017-02-07
Make all examples build on windows

OracPrime:
Make all examples build on windows by using cross-env before all explicitly path-referenced node commands


tamlyn:
Now the example are failing with `sh: cross-env: command not found`. Should this have been added as a dependency?

OracPrime:
Looking at package.json, I can see cross-env (^3.1.3) is listed as a devDependency.  Are you sure you did an npm install before attempting to rebuild?

PR #844, opened 2017-02-07
WIP React Intl + Next.js example app

ericf:


ericf:
This was added directly in the Next.js repo: https://github.com/zeit/next.js/tree/master/examples/with-react-intl

PR #851, opened 2017-02-14
Exporting hasLocaleData outside of library

DomBlack:
This will allow calling code to use hasLocaleData to see if the locale is supported. For instance if they have only loaded "en" but the browser reports "en-US" it prevents the user calling the same split and loop commands

### DIFF
--- a/src/react-intl.js
+++ b/src/react-intl.js
@@ -5,11 +5,11 @@
  */
 
 import defaultLocaleData from './en';
-import {addLocaleData} from './locale-data-registry';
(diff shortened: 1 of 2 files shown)


ericf:
Just so I understand, where would you be calling this within your app code?

DomBlack:
This is being used in our server side code to detect if we've loaded the locale, or can support it from the incoming HTTP request and if not, we fallback to en-GB.

ericf:
An app's translated messages are usually the limiting factor. App's locales < React Intl's locales < All locales in world.

Are you also checking if your app has translators for the user's requested locale?

Issue #872, opened 2017-02-28
Add support for `tagName` and `className` props in React components


ipanasenko:
Hi, guys

I've submitted PR that adds support for `tagName` and `className` props.

This will allow better styling for components without wrapping them into another component

https://github.com/yahoo/react-intl/pull/873

ericf:
This has been discussed several other times and is not something we'll be supporting. Sorry that you took the time to make a PR 😢

ipanasenko:
No problem, just couldn't find original discussion

PR #873, opened 2017-02-28
Add support for `tagName` and `className` props in React components

ipanasenko:
Resolves https://github.com/yahoo/react-intl/issues/872


ericf:
We don't plan to support this, and it has been brought up before.

PR #878, opened 2017-03-04
513 rich text formatting

ghostd:
This a first proposal for the rich text formatting (#513).

This implementation allows to replace some tags as described by Eric.
Eg: Your <x:0>email</x:0> has been sent.

This implementation has some drawbacks:
- the nested tags with the same name are not supported
- the self-closing tags are not supported

The previous React component interpolation does not change. I re-used the 'values' property to make the migration easier and to not change the sCU implementation.

The tags are parsed with RegExp and capturing groups.
I only tested the code against Firefox and Chrome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ lib/
 locale-data/
 node_modules/
 src/en.js
+
+# IntelliJ
+.idea/
--- a/src/components/message.js
+++ b/src/components/message.js
@@ -11,6 +11,7 @@ import {
     shallowEquals,
     shouldIntlComponentUpdate,
 } from '../utils';
+import {replacePlaceholders} from '../placeholders';
 
 export default class FormattedMessage extends Component {
     static displayName = 'FormattedMessage';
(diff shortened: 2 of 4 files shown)


yahoocla:

ghostd:
CLA signed :)

ghostd:
It was a POC to discuss about an implementation of #513. We can close it if someone has another implementation or we don't want this kind of complexity in the project.

tricoder42:
Hey @ghostd, I have similar but slightly different implementation in [LinguiJS](https://github.com/lingui/js-lingui). I use just indexes and the low-level api looks like this:

```
formatElements(
          "Before \n<0>Inside <1>\nNested</1>\n Between <2/> After</0>",
          [<a href="/about" />, <strong />, <br />]
)
```

which results in:

```
Before <a href="/about">Inside <strong>Nested</strong> Between <br/> After</a>
```

I use this function internally *after* I get the translation.

- nested tags are supported (even with the same name)
- self-closing tags are supported
- built-in elements and React components works just fine
- the only drawback I know if you want to use e.g. `<br />` tag more than once, you have to create a new index for each. Not a problem if the parameters are generated by macro, but it's annoying if you write it by hand.

Feel free to use any part of it if you find it useful: [source](https://github.com/lingui/js-lingui/blob/master/packages/react/src/format.js), [tests](https://github.com/lingui/js-lingui/blob/master/packages/react/src/format.test.js)

redonkulus:
Closing per https://github.com/formatjs/react-intl/pull/878#issuecomment-424777548. Can be re-opened if others find is valuable.

PR #880, opened 2017-03-04
613 custom error handling

ghostd:
The default handler property acts as previously: it uses `console.error` if the environment is not 'production'. Now the developer can override this setting and provides its custom error handler.

For now, the `createError` function returns a string (matching the previous behavior)... maybe we could return an `Error` object?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ lib/
 locale-data/
 node_modules/
 src/en.js
+
+# IntelliJ
+.idea/
--- a/src/components/provider.js
+++ b/src/components/provider.js
@@ -11,7 +11,12 @@ import IntlRelativeFormat from 'intl-relativeformat';
 import IntlPluralFormat from '../plural';
 import memoizeIntlConstructor from 'intl-format-cache';
 import invariant from 'invariant';
-import {shouldIntlComponentUpdate, filterProps} from '../utils';
+import {
+  createError,
+  defaultErrorHandler,
+  shouldIntlComponentUpdate,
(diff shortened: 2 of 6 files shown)


ggauravr:
@ghostd This is a great approach. However, I was wondering if instead of having to call `onError(createError(message))` every time, how about one function like `handleError`, which actually does what `createError` does, and calls `onError` handler.

Something like:

```js
export function handleError(message, exception) {
  const errorMsg = (exception) ? `\n${exception}` : '';
  const formattedErrorMsg = `[React Intl] ${message}${errorMsg}`;

  onError(formattedErrorMsg);
}
```

So, in places where we handle errors, we can just say `handleError(<msg>, e)` rather than the more verbose `onError(createError())`.

fes300:
any news on this? It would be very helpful to me :+1:

redonkulus:
@ghostd I'm ok with this approach. I guess the one downside is that it increases the bundle size since webpack/rollup used to remove the messages for production builds. Do you know how much it increases the size?

Secondly, can you resolve the conflicts?

ghostd:
Hi,

I rebased the branch.

I don't know how many are added by those error strings. The size of production bundle is 45293 bytes, and the size of the development bundle is 130391.

ghostd:
On the master branch i get:
* react-intl.min.js 44379
* react-intl.js 130009

redonkulus:
Thanks, so thats almost a kb, pretty small. I will take another pass at this before merging.

ghostd:
Done

PR #901, opened 2017-03-15
[WIP] Safely use context

maxdeviant:
A little background: our team is using MobX as our state management solution. We are also using `react-intl` for handling our i18n/l10n needs.

The problem we are facing is that when we change the current language and update the messages accordingly, some components in our component tree do not update with the new messages.

The issue appears to be that the context updates are being blocked by a parent component returning `false` on `shouldComponentUpdate`. In our case, the `observer` HoC from `mobx-react` is returning `false` in its `shouldComponentUpdate` since the MobX state has not changed. This prevents the subtree from re-rendering which means the `react-intl` components can not re-render in response to the context change.

You can see the issue depicted in this diagram (taken from the soon-to-be-mentioned article):

@mweststrate has a Medium post on [How to safely use React context](https://medium.com/@mweststrate/how-to-safely-use-react-context-b7e343eff076#.erav4i99t) in which he details the problem and also offers a potential solution.

I have added a failing test that shows the issue.

If this is a direction that `react-intl` wants to pursue, then I will work on getting the necessary changes made in this PR. If not, then our team will probably have to drop `react-intl` in favor of something else.

### DIFF
--- a/test/unit/components/message.js
+++ b/test/unit/components/message.js
@@ -115,6 +115,39 @@ describe('<FormattedMessage>', () => {
         expect(renderedOne).toNotBe(renderedTwo);
     });
 
+    it('should re-render when context changes and blocked by parent sCU', () => {
+        intlProvider = new IntlProvider({locale: 'en', defaultLocale: 'en'}, {});
+
+        class IronDome extends React.Component {
+            shouldComponentUpdate() {
+                return false;
+            }
+
+            render() {
+                return React.Children.only(this.props.children);
+            }
+        }
+
+        renderer.render(
+            <IronDome>
+                <FormattedMessage id="hello" defaultMessage="Hello, World!" />
+            </IronDome>,
+            intlProvider.getChildContext()
+        );
+        const renderedOne = renderer.getRenderOutput();
+
+        intlProvider = new IntlProvider({locale: 'en-US', defaultLocale: 'en-US'}, {});
+        renderer.render(
+            <IronDome>
+                <FormattedMessage id="hello" defaultMessage="Hello, World!" />
+            </IronDome>,
+            intlProvider.getChildContext()
+        );
+        const renderedTwo = renderer.getRenderOutput();
+
+        expect(renderedOne).toNotBe(renderedTwo);
+    });
+
     it('accepts `values` prop', () => {
         const {intl} = intlProvider.getChildContext();
         const descriptor = {


yahoocla:

maxdeviant:
@yahoocla CLA has been signed.

ghostd:
@maxdeviant the medium post mentions a library: https://github.com/mjackson/react-broadcast
did you try it?

Also, did you try injectIntl(YourComponent)?

maxdeviant:
@ghostd Well the `react-broadcast` library would have to be implemented in `react-intl`, since that is what is providing the context.

I have not tried using `injectIntl`, but I have noticed previously that wrapping any component with `injectIntl` basically voids the performance optimizations that MobX provides out of the box, so the components re-render every single time.

langpavel:
If you want use `shouldComponentUpdate` it is your responsibility for using it correctly.
One workaround to this can be use of [`react-deep-force-update`](https://github.com/gaearon/react-deep-force-update#react-deep-force-update) like I do [here](https://github.com/kriasoft/react-starter-kit/blob/feature/react-intl/src/components/App.js#L76). It is probably not the best way.
Better way is patch your `shouldComponentUpdate`, which is not easy: facebook/react#2517

maxdeviant:
@langpavel I do not implement `shouldComponentUpdate` manually, it's implemented in MobX, so it's not really possible to override/implement "correctly".

Like I said, if this is something that the library maintainers don't want to deal with, then so be it.

langpavel:
@maxdeviant Ok, implementing `shouldComponentUpdate` correctly is not possible by this way, my bad. BTW, great article, I just read that.
They suggest to create separate store in context for `IntlProvider` where other components will be subscribed, so updating `IntlProvider` will trigger explicit rerendering of explicitly subscribed components..
It looks nice, but for me, simpler is just use `react-deep-force-update` after you change language. It also makes sense to open issue on `react-deep-force-update` to traverse all components, but selectively mark only some (by predicate) for rerendering.

maxdeviant:
@langpavel I actually have a somewhat hacky workaround implemented at the moment.

Basically I created a MobX-aware provider that manages the translation state, and then I wrapped `FormattedMessage` with an `observer` that reacts to the language changes. It seems to be working so far, and it may do the trick long-term.

ghostd:
@maxdeviant i think your performance issue with injectIntl is the same as reported by #818 (not patched yet)

ChrisJamesC:
I have created a small library for combining react-intl and mobx. See https://github.com/Sqooba/mobx-react-intl. Maybe you can solve your issue by directly using it. 

When I faced this issue I solved it by passing the locale as key in the IntlProvider component. This ensured that all components bellow it were re-rendered when the locale (and the messages) were changed. 

For example: 
```
const {locale, messages} = this.props.localeStore; 
return <IntlProvider key={locale} locale={locale} messages={messages}>
    <App />
</IntlProvider>
```

PR #918, opened 2017-04-08
React 15.5 compatibility

inukshuk:
Thanks for `react-intl`! This PR just imports `PropTypes` from 'prop-types' package as suggested by React 15.5 and later.


inukshuk:
I also updated the examples to use 'prop-types' and changed the minimum peer dependency of React to version 0.14.9 (which is compatible with the 'prop-types' module).

I was a bit puzzled how to make the tests work with both React 14 and 15.5+ without deprecation messages. ES6 import statements must be static and at the top-level so we cannot conditionally import `createRenderer` from either 'react-addons-test-utils' (for React 14) or 'react-test-renderer/shallow' (for React 15). For the time being I solved this by creating 'test/renderer.js' in an NPM pretest script to make sure all tests continue to work for the `react:14` and `react:15` targets but I'd be happy if anyone could suggest a better solution. (Or if the react:14 tests can be dropped)

examples/nested/src/client/components/app.js
```diff
@@ -1,4 +1,5 @@
-import React, {Component, PropTypes} from 'react';
+import React, {Component} from 'react';
+import {PropTypes} from 'prop-types';
```
antoinerousseau: why brackets?
https://github.com/reactjs/prop-types#importing

inukshuk: Good question! I must be going blind.

bjankord:
@inukshuk This was a [solution](https://github.com/paypal/glamorous/blob/de50b7ccaeb37f60128b46b1d9d9d56cd474c35a/src/react-compat.js) I saw from @kentcdodds to handle importing different packages based on what version of react was used. This idea could be used to conditionally import `createRenderer` from either 'react-addons-test-utils' (for React 14) or 'react-test-renderer/shallow' (for React 15)

inukshuk:
@bjankord cool thanks!

Are the tests ever compiled to something other than commonjs modules? If that's not a requirement we can just use `require` directly -- that's much better!

inukshuk:
Alright, followed @bjankord's suggestion and used `require` directly; I think this should cover all compatibility issues now.

probablyup:
@ericf could you take a look at this please? FB shows console errors for these deprecations and it adds gross noise when trying to develop / debug

eMarek:
Any plans when this will be merged or have I missed something that we are waiting for?

MichaelDeBoey:
@ericf I think this can be merged?

tswaters:
Last commit from @ericf was almost a month ago, updating contributing docs for posting a new release.

Also there's this:  https://ericf.me/apple.  I'm not sure if ericf is working on this project any longer and/or there are any other contributors.... most of the time I see someone with a contributor badge it's ericf.

MichaelDeBoey:
@antoinerousseau is the reviewer, so maybe he can merge it then?
Or maybe @ericf knows who's following up this project in this place?

probablyup:
I wouldn't hold your breath for him to respond, tbh. Yahoo is all but dead at this point, would be nice if the FB open source team might be interested in taking it over? @gaeron?

MichaelDeBoey:
That could indeed be another possibility @probablyup :-)

gaearon:
We don’t really “take over” projects from other companies—and we also don’t maintain anything that we aren’t actively using in production. We have our own internalization infra at Facebook so it wouldn’t make sense for us to steer `react-intl`. If there is a need for a new maintainer, perhaps you could raise an issue? I’m sure some of the large companies using it could volunteer to help!

okuryu:
Just the other day I got a commit bit for react-intl from @ericf and @redonkulus (from Yahoo) to maintain react-intl. I have told them I would like to help this project. I think this change is a high priority for us, but please give me a little time to review this change and prepare to release a new version.

All feedbacks about this change are always welcome, please post your thoughts if you have any concern or idea.

bjankord:
Thanks for the update @okuryu!

MichaelDeBoey:
Any news on this one @okuryu? :-)

lyushenko:
We are slightly blocked by this issue and I'd like to contribute if you need any help @okuryu

bjankord:
Any update on this? @ericf @okuryu This is the last dependency I need updated to have react 15.5 compatibility before I can update to react 15.5.

mikebridge:
This is a trivial PR that remains unmerged after three weeks.   Is this project still active or is it a victim of Yahoo's collapse?

bmv437:
@mikebridge please read this comments above. Specifically [this one](https://github.com/yahoo/react-intl/pull/918#issuecomment-296641941). I expect @okuryu will merge this PR soon and cut a new release.

mikebridge:
@bmv437 Thanks, I have read the comments which is why I'm concerned.  I've only just started using react-intl, and I now get the impression that the core team has suddenly left the project—is this not correct?

okuryu:
Okay, I've merged. I'll release a new version within a week.

gaearon:
Just to be clear, this is really more of a 16 than 15.5 compatibility. Your apps will work even with warnings—it’s nice to gradually fix them as we move towards 16, but I think it’s also unnecessary to exert so much pressure on library maintainers right now. This change is not urgent.

robcaldecott:
Understood 100% but as this warning takes the form of a console error it causes tests to fail. Does anyone know of a good workaround for this?

lyushenko:
@robcaldecott you can use the production version of React to suppress console errors or install a fork temporarily:
```
"react-intl": "inukshuk/react-intl"
```
However, it would be better if React logged warnings instead of errors (`console.warn`)

probablyup:
Agreed. I raised a PR about that in fbjs and they rejected it.

robcaldecott:
I use create-react-app and that does not come with Jest configured to use the production version of React. I'll try setting NODE_ENV and see if that works.

gaearon:
I don't think `console.error` causes tests to fail in Create React App. What have you that impression?

gaearon:
(If it does cause CRA tests to fail please raise an issue with CRA. This was never intended. In general when you see weird behavior please raise issues rather than silently accept it 😉 )

robcaldecott:
On Windows with CI=true it returns an ERRORLEVEL of 1. I'll raise a PR. Weirdly it does not do this if you pass -- --coverage to npm test.

lyushenko:
@gaearon in our case, we're using Capybara to run feature tests and it is set up to capture all JS errors and our tests are failing due to deprecation warning. No big deal though, we kinda "hacked" it :)

gaearon:
Interesting. Please do raise it, thanks!

Issue #557, opened 2016-07-30
Alternative way to supply default messages?


tomdmaguire:
Hi there,

In the [react-intl v2 changelog](https://github.com/yahoo/react-intl/issues/162) regarding the `IntlProvider` component, it states:

> Note: How there's no defaultMessages prop, that's because it's assumed the default message descriptors live co-located to where the messages are being formatted.

I currently am storing all translations as a json files (one file per locale e.g. en-US.json, de.json) which then are then loaded and passed into the `messages` prop of the IntlProvider.

Ideally I do not want to supply a `defaultMessage` prop each time I render a `FormattedMessage` component, as I have all of my "default messages" in a en-US.json translation file. (I always want to fallback to this locale).

To to cater for my scenario I built a wrapper for the `FormattedMessage` component which is basically as follows:

``` js
import { FormattedMessage } from 'react-intl';
const defaults = require('en-US.json');

export default function Translate({ id, ...values }) {
  return (
    <FormattedMessage id={id} values={values} defaultMessage={defaults[id]} />
  );
}
```

This works ok until I am in a situation where I cannot use the `Translate` React component and instead need to use the [inject api](https://github.com/yahoo/react-intl/wiki/API#injection-api), as then I have to duplicate the loading default messages logic (`defaultMessage={defaults[id]}`), in all places where I use the injected `intl` obj.

Do you have any examples of solutions I could use to overcome this duplication and simply use a message id in scenarios where I need to translate string and somehow fallback to en-US if the translation does not exist in the current locale?

Any help is much appreciated!
Thanks,
Tom

ericf:
It's probably best if we add back the option to babel-plugin-react-intl to remove the extracted data. The issue was people were running the plugin multiple times by accident and getting errors. 

If you already have all of the default messages defined via the `defaultMessage` prop on FormattedMessage or using `defineMessages`, then you can omit providing the same messages to IntlProvider.

tomdmaguire:
FYI, I eventually decided on the following method to handle locale fallbacks:

```js
import locales from 'lib/locales';

const messages = language === 'en' // english is our default locale
  ? locales['en']
  : merge({}, locales[language], locales['en']); 

return (
  <IntlProvider
    locale={language}
    messages={messages}
  >
    {children}
  </IntlProvider>
);
```

derwaldgeist:
[...]

PR #925, opened 2017-04-13
Issue: https://github.com/yahoo/react-intl/issues/557

jianping-roth:
Problem: we have a number of language packages, when a key is missing from one language, we want to default to the corresponding English value. This is the default behaviour for GWT and Cocoon, the two web development frameworks we used previously.

The changeset implments this default behaviour. I added one more property called defaultMessages.

### DIFF
--- a/src/components/provider.js
+++ b/src/components/provider.js
@@ -27,6 +27,7 @@ const defaultProps = {
 
     defaultLocale : 'en',
     defaultFormats: {},
+    defaultMessages: {},
 };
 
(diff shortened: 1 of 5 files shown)


yahoocla:

jianping-roth:
Done. Jianping

redonkulus:
Will https://github.com/yahoo/react-intl/pull/1171 help you?

redonkulus:
Please reopen if needed.

PR #816, opened 2017-01-15
Improve build

ericf:
This upgrades to `rollup@0.41.4` which fixes issues with ES3 Babel transforms and Babel helpers which has reduced the build file size.

File | Before | After | Delta
---- | ------ | ----- | -----
react-intl.min.js | 47.10 kb | 41.61 kb | -11.66%
react-intl.min.js.gz | 13.31 kb | 11.51 kb | -13.52%

Fixes #814
Relates #788, #687

PR #930, opened 2017-04-20
Undo (soft) breaking change introduced in 93c0a35 & #816.

indexzero:
We say "soft" breaking because it was not a strict contract between `react-intl` and
consumers that raw `src/*` files be included in the npm package. If one were to want
a custom build of `react-intl` for their own purposes the npm package is now no longer
a viable location for that source code.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,6 @@
 coverage/
 examples/
 scripts/
-src/
 test/
 rollup.config.*
 .*


yahoocla:

Swaagie:
@indexzero https://github.com/yahoo/react-intl/blob/master/package.json#L34 is exposed as ES6 from the npm package though. So the issue might be as simple as getting the builder to pickup on that over `main`

indexzero:
There is more complexity here. The logic behind a custom build generally centers around avoiding additional calls to `addLocaleData`. Even though there is an ES6 main now, that is a `rollup` based build, so there is no way to ignore `src/en.js` if you do not wish to include the default locale data.

Overall for the `react-intl` maintainers who will eventually read this: removing the source files from an `npm` package is a breaking change because the implicit contract you were making before was "hey, here's the raw individual source files if you want to use them directly". That contract was broken in #816. Overall those src files are only an additional 12k in your `npm` package:

```
tar -czvf src.tgz src/
ls -al src.tgz
-rw-r--r--  1 [redacted]  11673 Apr 21 13:35 src.tgz
```

... so while I can appreciate the desire for a smaller _build_ this change has very little impact on anyone w.r.t. number of bytes downloaded.

okuryu:
Okay, let's merge this.

PR #934, opened 2017-04-27
fix bug: message is empty string cause error

greatbody:
fix bug: message is empty string cause error.
may cause unnessary error when we intent to have empty value as translate result

### DIFF
--- a/src/format.js
+++ b/src/format.js
@@ -200,7 +200,7 @@ export function formatMessage(config, state, messageDescriptor = {}, values = {}
 
     let formattedMessage;
 
-    if (message) {
+    if (Object.prototype.toString.apply(message) === '[object String]') {
         try {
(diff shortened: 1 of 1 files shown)


yahoocla:

okuryu:
I'm not sure about this fixes. Could you explain more details?

greatbody:
sorry for the PR. 
I wonder if there is a way to set locale specific translation result to empty string.
say:
zh.js:
export default {
  xxx: '',
}
en.js:
export default {
  xxx: 'bbb',
}

I have no idear how to solve this problem,that's why this PR comes.

papasmile:
@greatbody maybe you could try that suggestion posted at https://github.com/yahoo/react-intl/pull/488#issuecomment-224945705

redonkulus:
Please reopen if still needed.

PR #937, opened 2017-05-04
Add `postinstall-build` to enable npm install from git repos

bjrn:
Using [postinstall-build](https://github.com/exogen/postinstall-build) which is a helper for conditionally building an npm-compatible package on `postinstall`. It checks for the existence of a `/lib` folder (which would be the case if installing from npm).


yahoocla:

redonkulus:
@bjrn looks like npm supports this natively now via `prepare`. Check the readme https://github.com/exogen/postinstall-build/

Secondly, I’m missing the reason for this change. Why is it needed?

PR #990, opened 2017-07-15
allow object for description prop

mattkime:
Fixing prop definitions to match https://github.com/yahoo/react-intl/wiki/Components#message-descriptor

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -6,7 +6,7 @@
 
 import PropTypes from 'prop-types';
 
-const {bool, number, string, func, object, oneOf, shape, any} = PropTypes;
+const {bool, number, string, func, object, oneOf, shape, any, oneOfType} = PropTypes;
 const localeMatcher = oneOf(['best fit', 'lookup']);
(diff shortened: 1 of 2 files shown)


yahoocla:

mattkime:
license is signed.

eilinora:
+1 for getting this merged, makes testing harder with the added noise this causes. Would appreciate this being merged!

okuryu:
@papasmile  Thanks for your suggestions.

> * Contribution Guide states "If you are adding new functionality, or fixing a bug, provide test coverage." [2] You could a test to test/unit/components/message.js by updating or copying 'renders a formatted message in a '[3] and passing in a description which is an object, then expect console not to report any warnings. They've done similarly in a different test in that same file [4].

Sure. PR for `CONTRIBUTING.md` welcome.

> * very minor: linted, but it looks like convention is to add a space after comma inside an array.

I think a space after comma in an array is a common practice in JavaScript. We also already run linting by ESLint, but lint rules change is welcome if we need a little work.

And this change looks good to me. Let merge in this month if no concern by others.

PR #1011, opened 2017-08-18
Update CLDR to 31

ykzts:
ref https://github.com/yahoo/formatjs-extract-cldr-data/pull/13

> In CLDR v30 and v31, many Japanese sentences are corrected.


ykzts:
@okuryu Can I get a new release?

okuryu:
yup. probably this month.

PR #1017, opened 2017-08-30
Template for GitHub Issues

okuryu:
Often the reported issue is not a problem of `react-intl`. It may be a problem with intl.js polyfill or Unicode CLDR. I hope this template should be helpful to author of issues.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,20 @@
+**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
+
+<!--
+If this is a BUG REPORT, please:
+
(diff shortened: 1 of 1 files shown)


okovpashko:
@okuryu is there any reason not to require a bin with a reproducible example for issues considered as bugs?

okuryu:
if all reporter provide the minimum repro step, I thought that bin isn't necessarily required. It might be a bug on server side. Of course, if we can reproduce with bin, I think that bin is also good.

redonkulus:
@okuryu lerna also has a nice one you get ideas from https://github.com/lerna/lerna/blob/master/.github/ISSUE_TEMPLATE.md

okuryu:
Okay, I upgrade the template based on the good point of lerna. 😎

PR #1023, opened 2017-09-02
Support React v16

ykzts:
```
...
warning "react-intl@2.3.0" has incorrect peer dependency "react@^0.14.9 || ^15.0.0".
...
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "invariant": "^2.1.1"
   },
   "peerDependencies": {
-    "react": "^0.14.9 || ^15.0.0",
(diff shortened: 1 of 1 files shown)


okuryu:
I'm not familiar with React 16, but it seems to have several breaking changes. We don't need to care for them?
https://github.com/facebook/react/issues/10294

ykzts:
There are lots of internal changes in React v16. However, there are few changes to the public API. Backwards compatibility is maintained.

I have been using React Int16 for a while with React v16, but there was no problem.

PR #1035, opened 2017-09-22
Adapt hello-world example

tobiaslins:
I was searching for something like this for hours to get this done.
I wanted to write "one" or "no" instead of the numbers. 
It was hard to find in the documentation that `#` will replace it with the current value

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,9 +53,10 @@ class App extends Component {
             <p>
                 <FormattedMessage
                     id="welcome"
-                    defaultMessage={`Hello {name}, you have {unreadCount, number} {unreadCount, plural,
-                      one {message}
-                      other {messages}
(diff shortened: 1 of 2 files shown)


papasmile:
There's already a reference in wiki https://github.com/yahoo/react-intl/wiki/API#string-formatting-apis referring to Format.js message syntax https://formatjs.io/guides/message-syntax/  

"{plural} Format" section pretty much has your exact same code example.  Also what if someone wanted an example of using `number` (which you've removed)?  How about just updating hellow-world/README.md with a simple note like "For a full description of pluralization features, check out https://formatjs.io/guides/message-syntax/"

redonkulus:
Closing based on the recommendation from papasmile.

PR #1040, opened 2017-09-29
Replace string ref in injectIntl HOC with functional one

Panya:
We are using `injectIntl` with `{ withRef: true }` and in React 16 we get an error (e.g. "Element ref was specified as a string (wrappedInstance) but no owner was set. You may have multiple copies of React loaded.", see https://github.com/facebook/react/blob/master/docs/warnings/refs-must-have-owner.md) when using it with SSR. There is a workaround for this issue in React repo (see a discussion at https://github.com/facebook/react/issues/3344). This PR is trying to fix the issue by replacing string ref by functional one.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -41,7 +41,7 @@ export default function injectIntl(WrappedComponent, options = {}) {
           '`injectIntl()`'
       );
 
-      return this.refs.wrappedInstance;
+      return this._wrappedInstance;
     }
 
     render() {
@@ -49,7 +49,7 @@ export default function injectIntl(WrappedComponent, options = {}) {
         <WrappedComponent
           {...this.props}
           {...{[intlPropName]: this.context.intl}}
-          ref={withRef ? 'wrappedInstance' : null}
+          ref={withRef ? (ref => this._wrappedInstance = ref) : null}
         />
       );
     }


yahoocla:

Panya:
I signed the CLA.

bradbirnbaum:
+1 - I need this in order to upgrade to React 16

githubdoramon:
Is this ever going to be fixed?

StorytellerCZ:
@githubdoramon @Panya Let's start by fixing the tests.

Panya:
@StorytellerCZ well, let's start by accepting my CLA. I guess this project is unmaintained currently.

StorytellerCZ:
@Panya I'm working on my own version. Feel free to join: https://github.com/Freedombase/react-intl/pull/2

githubdoramon:
@StorytellerCZ how to use your version directly from npm?

StorytellerCZ:
@githubdoramon 
https://www.npmjs.com/package/@freedombase/react-intl

Remove the current `react-intl` and add `@freedombase/react-intl`. You will have to change the import statements as well.

redonkulus:
Changing the string refs could break consumers that rely on the ref in their application. I believe this change would require a major version bump. AFAIK, this issue does not prevent you from upgrading your application. We have updated many applications to React 16 internally without this issue.

ptomasroos:
Can you please explore how this could be a breaking behaviour @redonkulus ?

redonkulus:
@ptomasroos I was referring to changing the code from a string reference. If applications are relying on this, then changing it may break them. Unless I am mistaken about the use of this reference, please let me know.

ptomasroos:
How would a calling application be able to use the string reference or even tell a difference?
I thought the reason it was encapsulated in the method getWrappedInstance was just in order not to bleed details about the implementation.

PR #1064, opened 2017-11-13
Hoist statics in injectIntl

ArnoSaine:

PR #1073, opened 2017-12-12
Copies statics from the Wrapped component to the Injected component

Sparragus:
## Description
Copies static values set on the Wrapped component over to the Injected component.

Here are two examples where this is done: [`react-router`](https://github.com/ReactTraining/react-router/blob/469fc9f7f0077fd8c1d794ab93d519adbfa851ea/packages/react-router/modules/withRouter.js#L25) and [`react-redux`](https://github.com/reactjs/react-redux/blob/ea53d6fb076359a864a1d543568d951d4b3eab3d/src/components/connect.js#L365).

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,14 +64,15 @@
     }
   },
   "dependencies": {
+    "hoist-non-react-statics": "^2.3.1",
     "intl-format-cache": "^2.0.5",
     "intl-messageformat": "^2.1.0",
     "intl-relativeformat": "^2.0.0",
     "invariant": "^2.1.1"
   },
   "peerDependencies": {
-    "react": "^0.14.9 || ^15.0.0 || ^16.0.0",
-    "prop-types": "^15.5.4"
+    "prop-types": "^15.5.4",
+    "react": "^0.14.9 || ^15.0.0 || ^16.0.0"
   },
   "devDependencies": {
     "babel-cli": "^6.2.0",
--- a/src/inject.js
+++ b/src/inject.js
@@ -11,6 +11,7 @@ import React, {Component} from 'react';
 import invariant from 'invariant';
 import {intlShape} from './types';
 import {invariantIntlContext} from './utils';
+import hoistStatics from 'hoist-non-react-statics';
 
 function getDisplayName(Component) {
   return Component.displayName || Component.name || 'Component';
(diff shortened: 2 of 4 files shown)


Sparragus:
I know this is the third PR with this feature (see https://github.com/yahoo/react-intl/pull/433 and https://github.com/yahoo/react-intl/pull/1064), but the other PRs have been either forgotten or some check did not pass.

I seriously hope to see this PR become merged. Without it, devs need to make dirty workarounds to copy static values.

ArmandoAssuncao:
this could be merged.

infiman:
Having `preload` static method in my components for my SSR implementation, but `injectIntl` is not hoisting it after wrapping them. Any chance this PR will be merged? Thanks in advance!

kyusu:
This would be very helpful indeed!

ericketts:
I've had to create a wrapper for `injectIntl` that does a call to `hoistNonReactStatic`... this should certainly be merged

adrienharnay:
Yes please

StorytellerCZ:
My branch has it merged (together with other PRs) if you really need this: https://github.com/Freedombase/react-intl

redonkulus:
@okuryu do you want to review and merge?

hank7444:
Why it still not be merged??

kelly-tock:
What is the alternative workaround if this is not merged?

ericketts:
@kelly-tock I've defined a file in my project, called `injectIntl.js` (some of this may not apply to you if you're not a flow user):

```javascript
import {injectIntl as baseInjectIntl, type IntlShape} from 'react-intl';
import hoistNonReactStatic from 'hoist-non-react-statics';

type TOpts = {intlPropName: string, withRef: boolean};

export default function injectIntl<Props, SanitizedProps: $Diff<Props, {intl: IntlShape}>>(
  BaseComponent: React.ComponentType<Props>,
  options?: TOpts
): React.ComponentType<SanitizedProps> {
  const WrappedComponent = baseInjectIntl(BaseComponent, options);
  hoistNonReactStatic(WrappedComponent, BaseComponent);
  return WrappedComponent;
}

export type {IntlShape} from 'react-intl';
```

so you just import that with `import injectIntl, {type IntlShape} from 'utils/injectIntl`, and use it how you'd use the built in fn.

redonkulus:
Closed via https://github.com/yahoo/react-intl/pull/1141

Issue #1079, opened 2017-12-21
Use default translations as failback when IntlProvider is missing


maciej-ka:
🙌 and ❤️ for your work on this project.

What if instead of showing error we would display default translations when a component is not wrapped in IntlProvider?

Scenario
---
Say, we have `components/App.js`
```
import React from 'react';
import {defineMessages, injectIntl} from 'react-intl';

const messages = defineMessages({
  title: {
    id: 'title',
    defaultMessage: 'title from intl default'
  }
});

const App = (props) => <div>{props.intl.formatMessage(messages.title)}</div>

export default injectIntl(App);
```

Current behaviour
---
At the moment, if it's not wrapped in `<IntlProvider>`, the result will be:
<img width="1008" alt="error" src="https://user-images.githubusercontent.com/5403694/34258244-63f12da0-e65d-11e7-9453-1b631f83b423.png">

Dream behavior
---
Display a component using default translations
and throw console warning about missing provider wrapper
<img width="258" alt="screen shot 2017-12-21 at 14 47 07" src="https://user-images.githubusercontent.com/5403694/34258351-d8c05d36-e65d-11e7-921c-5122a88ce736.png">

maciej-ka:
Scenario behind the change is having a component that is included few applications. If one of these applications is English language only and could use default translation of the component, at the moment, it's still forced to have IntlProvider.

torifat:
Hi @sourabh2k15, can you have a look at this issue?

sourabh2k15:
yeah sure, I'll look into this

PR #1080, opened 2017-12-21
spike: gracefull failback for missing provider

maciej-ka:
Spike for: https://github.com/yahoo/react-intl/issues/1079

This is just a fast example that when `IntlProvider` is missing, intl injected component could still work using default translations.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -10,7 +10,7 @@
 import React, {Component} from 'react';
 import invariant from 'invariant';
 import {intlShape} from './types';
-import {invariantIntlContext} from './utils';
+import IntlProvider from './components/provider';
 
 function getDisplayName(Component) {
   return Component.displayName || Component.name || 'Component';
@@ -30,7 +30,6 @@ export default function injectIntl(WrappedComponent, options = {}) {
 
     constructor(props, context) {
       super(props, context);
-      invariantIntlContext(context);
     }
 
     getWrappedInstance() {
@@ -45,12 +44,21 @@ export default function injectIntl(WrappedComponent, options = {}) {
     }
 
     render() {
-      return (
-        <WrappedComponent
-          {...this.props}
-          {...{[intlPropName]: this.context.intl}}
-          ref={withRef ? 'wrappedInstance' : null}
-        />
+      if (!this.context.intl) {
+        const DeepWrappedComponent = injectIntl(WrappedComponent);
+        return (
+          <IntlProvider>
+            <DeepWrappedComponent />
+          </IntlProvider>
+        );
+      }
+
+     return (
+       <WrappedComponent
+         {...this.props}
+         {...{[intlPropName]: this.context.intl}}
+         ref={withRef ? 'wrappedInstance' : null}
+       />
       );
     }
   }
--- a/test/unit/inject.js
+++ b/test/unit/inject.js
@@ -39,14 +39,6 @@ describe('injectIntl()', () => {
         });
     });
 
-    it('throws when <IntlProvider> is missing from ancestry', () => {
-        const Injected = injectIntl(Wrapped);
-
-        expect(() => renderer.render(<Injected />)).toThrow(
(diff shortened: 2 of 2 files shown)


yahoocla:

maciej-ka:
CLA signed

mediafreakch:
@maciej-ka Any chance you could finalize this PR? Make the licence check pass and fix the linting errors?

maciej-ka:
Thanks for looking at this. For a moment it's just a proof-of-concept, I'm still waiting for a green light in a related issue: https://github.com/yahoo/react-intl/issues/1079. Do we even want this?

mediafreakch:
I can only speak for myself: yes. Tried to find reasons why this check is there in the first place. Couldn't find any except preventing developers from their own misery 🙈

maciej-ka:
One real-life scenario has been added to issue description. I'll upgrade this spike to green status. Probably for the final version, I would need an advice about the direction this implementation should go.

maciej-ka:
CLA Signed, linter fixed, one test removed (that no IntlProvider should throw an error).

When this PR would get closer to merge a test will be added to check that default translation is displayed when there is no wrapping IntlProvider.

src/inject.js
```diff
@@ -45,12 +44,21 @@ export default function injectIntl(WrappedComponent, options = {}) {
...
 
     render() {
-      return (
-        <WrappedComponent
-          {...this.props}
-          {...{[intlPropName]: this.context.intl}}
-          ref={withRef ? 'wrappedInstance' : null}
-        />
+      if (!this.context.intl) {
+        const DeepWrappedComponent = injectIntl(WrappedComponent);
```
lucaas: We were using an approach similar to this in our application code and found that the component is recreated on each render if the intl context is missing.

I solved this by only creating the `DeepWrappedComponent` once and storing it on the `InjectIntl` instance.

jnak:
Hello,

This PR would be very helpful since it would allow to share components that have internationalized without forcing consumers to depend explicitly react-intl. Any update on this? Anything I can do to help this move forward?

Thanks,
J

redonkulus:
I'm not completely against this change but we need to ensure that the developer is aware of the issue that needs to be fixed. I don't want to hide the issue by using the defaultMessage. You will also need to add a test to assert the new behavior.

redonkulus:
Also, https://github.com/yahoo/react-intl/pull/1171 is another way to resolve this issue with I like as well.

jnak:
@redonkulus I also prefer https://github.com/yahoo/react-intl/pull/1171: 

- It does not require the component trees to be wrapped injectIntl(...) in order to avoid throwing. Component trees are normally able to use <FormattedMessage /> without calling injectIntl(...) as long as their parents are wrapped with an <IntlProvider />. I think it make sense to always fail gracefully when there is no <IntlProvider /> - wether injectIntl(...) was called on parent or not.

- It is a better from a separation of concern perspective to keep the fix internal to the <FormattedMessage /> since this is where the error was being thrown.

- The PR is mostly ready to be merged. It properly adds warning in dev. We just need to add one more test case.

Let's focus on https://github.com/yahoo/react-intl/pull/1171?

maciej-ka:
Sounds good. I posted this half year ago as fast spike and today barely remember what was it about.

sourabh2k15:
#1171 has been merged as of now, so does that eliminate the need for this PR?

jnak:
@sourabh2k15 Yes it makes sense to close this one for now. 

The only thing that this PR addresses that https://github.com/yahoo/react-intl/pull/1171 does not is adding graceful fallbacks for other <Formatted* /> components - not just <FormattedMessage />. But as per our discussions above and in the other thread, this PR is not the best approach to do this. 

If there is a need for graceful fallback of other <Formatted* /> components, we should send PRs similar to https://github.com/yahoo/react-intl/pull/1171 for each <Formatted* /> component.

PR #1093, opened 2018-01-22
Add option to config to let digits at west digit

samiashajrawi:
Add useWestDigits  to keep digits at west digits after converted to nonWest format. for example languages at below chart 
arabic_numerals-en svg

This changes keep local formats with west digits. which help users to have constant digits overall the web.

### DIFF
--- a/src/components/provider.js
+++ b/src/components/provider.js
@@ -28,6 +28,7 @@ const defaultProps = {
 
   defaultLocale: 'en',
   defaultFormats: {},
+  useWestDigits: false
 };
 
 export default class IntlProvider extends Component {
@@ -131,7 +132,7 @@ export default class IntlProvider extends Component {
         ...config,
         locale: defaultLocale,
         formats: defaultFormats,
-        messages: defaultProps.messages,
+        messages: defaultProps.messages
       };
     }
 
--- a/src/format.js
+++ b/src/format.js
@@ -29,6 +29,7 @@ const RELATIVE_FORMAT_THRESHOLDS = {
   month: 12, // months to year
 };
 
+
 function updateRelativeFormatThresholds(newThresholds) {
   const {thresholds} = IntlRelativeFormat;
   ({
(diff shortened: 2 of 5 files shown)


StorytellerCZ:
Would be nice to include other languages as well, like Chinese, Japanese and Cyrillic numerals.

juandopazo:
What's the motivation behind this change?

samiashajrawi:
@juandopazo to fix Like those issue https://github.com/yahoo/react-intl/issues/923

juandopazo:
I'm concerned this might lead to inconsistencies that are hard to understand when changing digits but not changing formatting rules.

Normally this is a kind of change that I'd prefer to see in the `Intl` API before doing it in `react-intl` as `react-intl` is designed to be a thin layer on top of the `Intl` API. However, since `Intl` takes years to get updates, I can understand the need to make changes here. I'd like to do a bit of research and see if the ICU libraries have tried anything like this or if there has been any discussions about it. I'll also try to get @ericf's feedback.

mgoenka:
@juandopazo: Is there any update on this? It does seem like using the western numerology system is a standard followed on a lot of websites, across the internet. Besides, this PR only introduces it as an option. The users of the package could choose to ignore the option.

caridy:
I don't think this is the right approach. Intl, and abstractions around Intl are expecting data as input, and produced UI elements that will make sense for each individual user. This PR changes that notion to also accepted abstract representations of the input data.

This library should remain as an abstraction for Intl, and expecting data in a similar fashion as Intl. It should only accept things that are valid data representation in javascript.

If you want to transform other representations of data into JS primitive values, you can have another library to do so, and call it right before calling react-intl to massage the data.

juandopazo:
@caridy the problem is that this is actually close to what `NumberFormat` does. The goal here is to keep formatting of the number, while not translating the digits. So, if the number in language FooBar is `Foo,BarBarBar`, the change in this PR keeps the comma and renders `1,000`. But if language YesNo formats numbers as `Yes.NoNoNo`, this PR would render `1.000`. So it's not a totally new way of representing data.

I understand expecting this to handle data in similar ways to the Intl API and this PR could rearrange its data to match its style better, but it kind of still needs to be a post-processing step. What do you think?

devcorpio:
Nice! my workmates and me are waiting with excitement this change!

caridy:
hmm, I think I got this wrong the first time, apologies. I think I understand now, let me try again:

What you want is to format something in a particular locale but using a different numbering system. This is perfectly supported today by browsers via extension keys, and we want to continue extending that extension mechanism. Here is an example:

```
new Intl.NumberFormat('en-u-nu-arab', {}).format(12345)
"١٢٬٣٤٥" 
```

As you can see, I'm formatting the number in locale: `en`, while forcing the system to use a different numbering system via the extension key `nu`, in this case setting that to `arab`. This extensions are specified in ICU, and supported via Intl today.

Of curse, this is not just about the digits, but the numbering system in general, including digits. Which mean that what you produce does makes sense. If you you want is to show them in `latn`, you can do this:

```
new Intl.NumberFormat('ar', {}).format(12345)
"١٢٬٣٤٥"
new Intl.NumberFormat('ar-u-nu-latn', {}).format(12345)
"12,345"
```

More details here: http://cldr.unicode.org/translation/numbering-systems and here: http://www.unicode.org/reports/tr35/#u_Extension

zbraniecki:
Yep, you should use numbering system (we're working on adding it to the options bag - https://github.com/tc39/ecma402/pull/175 )

juandopazo:
This is awesome!

Would a temporary hack like this make sense?

- Add a `numberingSystem` prop
- If `numberingSystem` is set, add a Unicode _u_ extension

samiashajrawi:
:+ That is awesome.  Thank you

devcorpio:
Hi,
Is that pull request merged?

mgoenka:
@devcorpio: Given @caridy's suggestion, it didn't seem like we needed this PR at all.

PR #1118, opened 2018-04-09
update `FormattedMessage` to accept an element as `tagName` prop

iDuuck:
I just searched for a way to supply a simple React component as `tagName`. Added the component as the `tagName` prop, which is also working, but raising an "Invalid prop types" error.

Don't know, if this is the desired behavior. If not, please just close the PR.

### DIFF
--- a/src/components/message.js
+++ b/src/components/message.js
@@ -23,7 +23,7 @@ export default class FormattedMessage extends Component {
   static propTypes = {
     ...messageDescriptorPropTypes,
     values: PropTypes.object,
-    tagName: PropTypes.string,
+    tagName: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
     children: PropTypes.func,
   };
 
--- a/test/unit/components/message.js
+++ b/test/unit/components/message.js
@@ -182,7 +182,7 @@ describe('<FormattedMessage>', () => {
         expect(renderedThree).toNotBe(renderedOne);
     });
 
-    it('accepts `tagName` prop', () => {
+    it('accepts string as `tagName` prop', () => {
         const {intl} = intlProvider.getChildContext();
         const descriptor = {
             id: 'hello',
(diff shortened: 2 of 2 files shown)


redonkulus:
@iDuuck Whats the use case where string isn't sufficient?

iDuuck:
For example in a Headline, it bugs me to have an extra `span` element in an `H1` tag.

redonkulus:
You can always use the API directly and not use the React component. https://github.com/yahoo/react-intl/wiki/API#formatmessage

chrismiceli:
@redonkulus, the api doesn't support jsx as a value so they are not exactly the same.

iDuuck:
I mean, what is the `tagName` property for, if not for something like that purpose?

adrian-gallardo-wizeline:
I was just looking for this functionality too.. I think it make sense to use a `React.Fragment` in `FormattedMessage` and let the user decide if they want a `span`, `div` or `none` (fragment) element to wrap the translation..
@redonkulus , about why trying to use a `React.Component` instead the function I think it is answered here: [Why Components?](https://github.com/yahoo/react-intl/wiki/Components#why-components)

- Render React elements that seamlessly compose with other React components.
- Support rich-text string/message formatting in <FormattedMessage>.
- Implement advanced features like <FormattedRelative>'s updating over time.
- Implement shouldComponentUpdate to avoid costly formatting operations.
- Provide Prop Type warnings for formatting options.

chrismiceli:
I am all for fragment option with the component so you can have no wrapping element.

jcready:
@adrian-gallardo-wizeline @chrismiceli you can have this already: https://github.com/yahoo/react-intl/wiki/Components#intlprovider

```jsx
<IntlProvider textComponent={React.Fragment}>
```

lexanth:
@redonkulus I don't think this PR is actually right - element is for an instance of a React element, not for the class/function to create it. See https://codesandbox.io/s/m5v371x409 - giving console warnings for exactly this. To cover classes/SFCs I think it should just be function and fragments are symbols (although you may want to be more precise).

lexanth:
@redonkulus No, pretty sure node is for an instance rather than a type again e.g. when they say string, they mean the string content that you would display, not an HTML tag name.

I updated that sandbox with an example of a PropType that does work, although it's a bit imprecise on Fragment (and what strings are valid tag names, for what it's worth) - https://codesandbox.io/s/m5v371x409

```
  tagName: PropTypes.oneOfType([
    PropTypes.func,
    PropTypes.string,
    PropTypes.symbol // Covers Fragment
  ])
```

Maybe one to raise on the prop-types repo, see what they suggest.

booc0mtaco:
@redonkulus @lexanth is this work continuing on another PR or issue ticket? I was troubleshooting a warning where a developer added `tagName={React.Fragment}`. It works for us, but adds the warning in the console as described above. 

(we're on 2.4.0 right now, but after testing a newer release with this change, it still has the prop type warning in this case)

PR #1121, opened 2018-04-16
Change `console.error()` to `console.warn()` for #1120

### LINKED ISSUE #1120: Default locale error instead of warning in React Native app
Hello! I've looked through issues, looks like it wasn't reported yet.

I have option to switch between locales, that have not been added to `IntlProvider` yet.

### Expected behavior
I want to get `YellowBox` message that react-intl is using `defaultLocale` as warning.

### Current behavior
I get runtime error and my app crashes. I believe that this is because this warning is reported with `console.error()` instead of `console.warn()`

I added `react-intl` with yarn and switched to locale I didn't add in debug mode
`react-native run-ios`

```js
export default class App extends Component {
  static defaultProps = {
    locale: 'ru' // comes from redux
  };

  render() {
    const {locale} = this.props;

    return (
      <IntlProvider
        textComponent={Text}
        locale={locale}
      >
        <View>
          <FormattedMessage id='Example' defaultMessage='Example'/>
          <AppRouter/>
        </View>
      </IntlProvider>
    );
  }
}
```

### Your Environment

Building 

| Executable | Version |
| ---: | :--- |
| `npm ls react-intl` | v2.4 |
| `npm ls react-native` | v0.54.1 |
| `npm --version` | 5.6.0 |
| `node --version` | 9.4.0 |

| OS | Version |
| --- | --- |
| macOS Sierra | 10.12.6 |

iphone 6 - ios 11 0 2018-04-16 17-22-53

naffiq:
To prevent React Native from crashing in dev environment without locale being loaded to `IntlProvider`
Closes #1120

### DIFF
--- a/src/components/provider.js
+++ b/src/components/provider.js
@@ -116,7 +116,7 @@ export default class IntlProvider extends Component {
       const {locale, defaultLocale, defaultFormats} = config;
 
       if (process.env.NODE_ENV !== 'production') {
-        console.error(
+        console.warn(
           `[React Intl] Missing locale data for locale: "${locale}". ` +
             `Using default locale: "${defaultLocale}" as fallback.`
         );
--- a/src/format.js
+++ b/src/format.js
@@ -47,7 +47,7 @@ function getNamedFormat(formats, type, name) {
   }
 
   if (process.env.NODE_ENV !== 'production') {
-    console.error(`[React Intl] No ${type} format named: ${name}`);
+    console.warn(`[React Intl] No ${type} format named: ${name}`);
   }
 }
 
@@ -67,7 +67,7 @@ export function formatDate(config, state, value, options = {}) {
     return state.getDateTimeFormat(locale, filteredOptions).format(date);
   } catch (e) {
     if (process.env.NODE_ENV !== 'production') {
-      console.error(`[React Intl] Error formatting date.\n${e}`);
+      console.warn(`[React Intl] Error formatting date.\n${e}`);
     }
   }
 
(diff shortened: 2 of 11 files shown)


yahoocla:

naffiq:
Oops, looks like this job failed
[https://travis-ci.org/yahoo/react-intl/jobs/367113188](https://travis-ci.org/yahoo/react-intl/jobs/367113188)
Although it's not related to my changes. I've found related test case: https://github.com/yahoo/react-intl/blob/master/test/unit/components/relative.js#L199
Maybe increase jasmine async timeout a little bit?

redonkulus:
The problem is that these are actually errors, seems weird to change the implementation for a specific library/environment. How do other libraries that work with React Native handle this? I can't foresee everyone switching to `console.warn`.

redonkulus:
> Maybe we can detect if textComponent prop is RN component to decide, whether it should error or warn?

What would this look like if done?

naffiq:
@redonkulus 

There is a hacky way, that relies on `React.View` – in React web apps it's undefined, while in RN projects it is not.

naffiq:
@AndrewIngram 

Now when I think of it, maybe it's better allowing developers overriding the method and adding 3rd party loggers, such as sentry to gather untranslated parts of UI and other warnings in production

ghostd:
This could be handled with #880

redonkulus:
So I've just merged https://github.com/yahoo/react-intl/pull/880, you can try this out as it gives you the ability to customize the error reporting in your application.

Issue #1106, opened 2018-03-03
RFC: New React Context API


langpavel:
### This is RFC

There is upcoming change in React 16.3: [New Version of Context](https://github.com/reactjs/rfcs/blob/master/text/0002-new-version-of-context.md#summary)

Is there any plan for new (*probably major / may be minor*) release of `react-intl` supporting this new API?
 - I believe that some issues can disappear with new API
 - old APIs will probably start throwing deprecation warnings with React 16.4
 - old APIs will probably be removed in React 17

iamstarkov:
> I believe that some issues can disappear with new API

For sure, for example `shouldComponentUpdate` wont anymore break language switch

iamstarkov:
i can drive this

iamstarkov:
@okuryu @ericf what do you think about it updating react-intl to a new context api?

redonkulus:
@iamstarkov can it be done in a backwards compatible way?

langpavel:
@redonkulus new Context API is `react>=16.3`, old context will be supported in `react<17`

iamstarkov:
@redonkulus yes, i crafted https://github.com/cssinjs/theming and ensured that it works properly for context updates

jantorestolsvik:
React 16.3 is out. Is this being worked on? I can help if you need.

louh:
Is the idea behind adopting the official Context API that we can dynamically re-render an app when the locale changes, without needing a `key` prop on the `IntlProvider`?

jantorestolsvik:
Yes, and we can rerender only the FormattedMessage etc components and not the whole tree

heathzipongo:
It would be very helpful if `react-intl` would export it's own React Context.Consumer component implementation so that I can access the `intl` object from inside my component without the need for the `injectIntl` HOC.  Please tell me that is also part of the work being done?!

i.e. in my component I can do something like:
```
import {IntlContextConsumer} from 'react-intl'; // Proposed name for the consumer
class MyComponent extends Component {
  render() {
    return (
      <IntlContextConsumer>
        {{intl} => {(
           <input 
             placeholder={intl.formatMessage({id: 'messageId', defaultMessage: 'Enter something'})
             ... // other props
          />
        }}
      </IntlContextConsumer>
    );
  }
}
```

langpavel:
@heathzipongo Yes, this is example case of new context API. I hope this will be supported in the future. And it should be possible to use intl outside of context at your own risc…
[...]

Issue #695, opened 2016-10-18
What actually happens if I change locale? Remount?


adjourn:
I accidentally noticed something and it got me wondering..

I have a component that animates up and down if boolean `state` changes. This component is **always** rendered in my app. However, if it's in non-default `state` and I change locale, it doesn't animate to default `state` or remain to same `state`, it just goes &#42;poof&#42; to it's original position. 

I concluded that it doesn't just get re-rendered, it gets remounted, otherwise it would animate or what?

Am I right or have I made a mistake somewhere? I would prefer to avoid that kind of jank. Could I?
## 

My theory: if this is a conscious choice, it's because `react-intl` uses `context` which doesn't automatically re-render components if it changes and we would have to write this to every component that uses `react-intl` (which also has some caveats):

``` js
shouldComponentUpdate(nextProps, nextState, nextContext) {
    return !shallowEqual(this.props, nextProps) ||
           !shallowEqual(this.state, nextState) ||
           !shallowEqual(this.context, nextContext)
  }
```

If Im right and this is intentional, it means that I have to choose language switcher location very carefully, because all `state` goes to default and it will make a jank in app flow (closes all modals, clears inputs _etc_).

unirey:
I found issue #234 after few hours of browsing and it that seems to describe my issue: I had `key={locale}` on `<IntlProvider>` and it forces full teardown.

 facebook/react#2517 is still open and doesn't seems to be resolved. I take that `key={locale}` is still our only workaround. If I understand the issue correctly, only other solution would be to expose all components to `context` and use `shouldComponentUpdate()` but that would lose the purpose of  `context` and probably had an impact to performance.

Does anybody has suggestions to make transition as smooth as possible? 

--
Few ideas:
- keep track of inputs, if user has filled any inputs, ask a confirmation before locale switch
- make a short (_e.g_ 500ms) fade-in-out overlay animation before actually switching locale

ericf:
Redux has this same issue, but they are working on ways around not having to implement `shouldComponentUpdate` on their `<Provider>`. Ideally we'd be able to rebuild `<IntlProvider>` in a similar fashion so it can be performant without `shouldComponentUpdate()`.

I don't currently have the time to do this, but would be up for working with someone who does.

Dmitry-N-Medvedev:
@ericf, what if the context is not used at all for locale switching? what if every component that depends on locale just listens for an locale switching "event" on redux store?

bhouser:
[...]

PR #1124, opened 2018-04-25
Add react-intl example using Redux.

EhabGamal:
# react-intl example using Redux
This is a simple and runnable example showing how to use React Intl with [Redux](https://github.com/reactjs/redux). And how to dispatch actions to change the locale language on the fly.

### DIFF
--- a/examples/redux/.babelrc
+++ b/examples/redux/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["react", "es2015", "stage-1"]
+}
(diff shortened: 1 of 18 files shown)


papasmile:
Hi, Guys: definitely appreciate any and all contributions via PR process.  Looking at proposed example code, it is using a questionable approach of modifying locale on `IntlProvider` as discussed in many issues including https://github.com/yahoo/react-intl/issues/695 .  

AFAIU community wants to rally around new context APIs via https://github.com/yahoo/react-intl/issues/1106 before settling on a recommended approach to changing locale.

Issue #811, opened 2017-01-09
injectIntl bug React.Component instance


marlonmleite:
I've implemented **ref="myRef"** in my component and it does not return an instance of the component.
It is returning an instance of InjectIntl ​​instead of the component.

This happens because of this:
`export default injectIntl(MyInput) or @injectIntl ​​annotation`.

I need to instantiate my component for actions and focus on my **input** and react-intl is disrupting this.

My entry code:
```
  render() {
    const {children, placeholder, intl, ...props} = this.props
    const formattedMessage = placeholder ? intl.formatMessage({id: placeholder}) : ''

    return(
      <div className="default">
        <input
          className="input"
          placeholder={formattedMessage}
          ref={(elem) => this.configInput(elem)}
          {...props}
        />
        <div className="err">{children}</div>
      </div>
    )
  }
```

ericf:
You should set the `withRef` option when calling `injectIntl()`: https://github.com/yahoo/react-intl/wiki/API#injection-api

This would allow you to access the wrapped instance via `getWrappedInstance()`. You keep your original `ref` to the inject intl wrapper component, and call `getWrappedInstance()` on it.

PR #1201, opened 2018-10-18
Make react-intl react 16 compatible

### LINKED ISSUE #1078: withRef only work before React 16
BUG REPORT 
### Expected behavior
getWrappedInstance should return the WrappedInstance ref

### Current behavior
getWrappedInstance is not returning the WrappedInstance ref

use injectIntl(..., { withRef: true });
### Your Environment
React 16

The problem is related that you are using a deprecated way of setting refs.
To fix it you should change in injectl.js:
-at line 52 change to `ref={withRef ? (wrappedInstance) => { this.wrappedInstance= wrappedInstance; } : null}`
-at line 44 change to `return this.wrappedInstance;`

ptomasroos:
Closes #1078
This is not a breaking change since the access of the ref is used when calling getWrappedInstance()

In order to support React 16. We need to remove the legacy way of using ref's by setting a string.
This is now marked as legacy.

```
We advise against it because string refs have some issues, are considered legacy, and are likely to be removed in one of the future releases. 

Note

If you’re currently using this.refs.textInput to access refs, we recommend using either the callback pattern or the createRef API.
```

https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -50,7 +50,7 @@ export default function injectIntl(WrappedComponent, options = {}) {
         <WrappedComponent
           {...this.props}
           {...{[intlPropName]: this.context.intl}}
-          ref={withRef ? (ref => this._wrappedInstance = ref) : null}
+          ref={withRef ? /* istanbul ignore next */ (ref => this._wrappedInstance = ref) : null}
         />
       );
     }
--- a/src/inject.js
+++ b/src/inject.js
@@ -42,7 +42,7 @@ export default function injectIntl(WrappedComponent, options = {}) {
           '`injectIntl()`'
       );
 
-      return this.refs.wrappedInstance;
+      return this._wrappedInstance;
     }
 
     render() {
@@ -50,7 +50,7 @@ export default function injectIntl(WrappedComponent, options = {}) {
         <WrappedComponent
           {...this.props}
           {...{[intlPropName]: this.context.intl}}
-          ref={withRef ? 'wrappedInstance' : null}
+          ref={withRef ? (ref => this._wrappedInstance = ref) : null}
         />
       );
     }


ptomasroos:
As mentioned by @ericf in this usage how to access the ref from the injectIntl component.
https://github.com/yahoo/react-intl/issues/811#issuecomment-271915903

And thats how we use it through out our web app's as well.

ptomasroos:
Tests are failing because of current state in master i would say?

```
Jest: Coverage for functions (98.73%) does not meet global threshold (100%)
```

redonkulus:
Thanks for the PR. I'm fine merging if it doesn't break anyone. 

As for the failing build, it looks like one of the tests are failing:

```
 FAIL  test/unit/react-intl-with-locales.js
  ● Test suite failed to run
    /home/travis/build/yahoo/react-intl/src/utils.js:2
    }});
    ^
    
    SyntaxError: Unexpected token }
      
      at new Script (vm.js:79:7)
      at Object.<anonymous> (src/inject.js:20:572)
```

It's not related to your PR, since master has the same problem, but do you mind seeing if it's an easy fix?

ptomasroos:
Sadly i can't reproduce this locally with same node version and can't really find the cause for this.
Seems to happen only on node 10 for some reason?

redonkulus:
So the unit tests pass but the coverage is below 100%. `master` is 100% for all functions and those builds are passing. Can you get the `inject` file back up to 100% coverage?

ptomasroos:
TBH looking at the tests it look seriously solid. @redonkulus

ptomasroos:
But i will take a look if I can find something

ptomasroos:
Sorry @redonkulus I cant bring this to 100, its complaining about the ref callback assignment and I can't really find a way to test that "logic". Can you give me some insights?

ptomasroos:
Thanks for the patience @redonkulus.
I gave it a try through several examples out on the web, yours as well. But couldn't make it to get covered.

Hence I added a ignore line instead since I already feel this is covered through the existing lines which check the ref assignment through the invariant checks.

Would be great if we could merge and release this in order to avoid splitting the community into branches and forks as currently 16.6 users need to use https://www.npmjs.com/package/@freedombase/react-intl

redonkulus:
published in v2.7.2

PR #1221, opened 2018-12-07
[fix] Compatibility of injectIntl and static contextType

ntucker:
React 16.6 introduces a new static `contextType`. If you use it in a component wrapped by injectIntl and old hoist-non-react-statics it will copy contextType to the wrapper component thus breaking context integration. Old versions were unaware of this so they would copy them through. React will then look at contextType and ignore contextTypes that injectIntl uses. It will then complain it can't find the context provided by react-intl.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     }
   },
   "dependencies": {
-    "hoist-non-react-statics": "^2.5.5",
+    "hoist-non-react-statics": "^3.2.1",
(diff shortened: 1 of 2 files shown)


yahoocla:

ntucker:
I signed the CLA but the check is still showing failed. How do I restart that check?

redonkulus:
Closing in favor of https://github.com/formatjs/react-intl/pull/1272

PR #1228, opened 2019-01-07
Fix PropType of Formatted{,HTML}Message to accept any component

wereHamster:
A React component is not the same as a React element! In our case, the tagName prop can be anything that can be passed to React.createElement(). Unfortunately the prop-types package doesn't provide a suitable validation function, so people are forced to build their own.

See https://github.com/facebook/prop-types/issues/200 for a related discussion.

### DIFF
--- a/src/components/html-message.js
+++ b/src/components/html-message.js
@@ -23,7 +23,11 @@ export default class FormattedHTMLMessage extends Component {
   static propTypes = {
     ...messageDescriptorPropTypes,
     values: PropTypes.object,
-    tagName: PropTypes.string,
+    tagName: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.func,
+      PropTypes.shape({render: PropTypes.func.isRequired}),
+    ]),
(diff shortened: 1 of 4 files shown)


yahoocla:

wereHamster:
@yahoocla please check again.

redonkulus:
ljharb mentions in that issue:
> "it has a render function" is a pretty brittle check, when react-is has a robust isForwardRef method available.

Should we try to leverage `react-is` as well?

wereHamster:
I didn't want to add another dependency. Ideally prop-types would provide that validation itself, once it does you can switch to it. Until that time, this is a net improvement IMO.

wereHamster:
Alright, I'll close this and wait until the prop-types package provides the validation. In the meantime, I'll be using this to silence the warning (I rely on TypeScript anyway):

```
if (process.env.NODE_ENV !== "production") {
  require("react-intl").FormattedMessage.propTypes.tagName = require("prop-types").any;
}
```

Issue #1229, opened 2019-01-08
Update formatjs-extract-cldr-data to latest version (2 major versions behind)


essensx:
Please update `formatjs-extract-cldr-data` package to latest version (6.0.0 as of writing).

Thanks!

okuryu:
I just created a PR #1230 for this.

okuryu:
Published `react-intl@2.8.0`. Thanks!

PR #1230, opened 2019-01-08
Update to CLDR 34

okuryu:
This upgrades the formatjs-extract-cldr-data from v4.0.0 to v6.0.0.

It includes following changes.

- CLDR 34: http://cldr.unicode.org/index/downloads/cldr-34
- CLDR 33: http://cldr.unicode.org/index/downloads/cldr-33
- CLDR 32: http://cldr.unicode.org/index/downloads/cldr-32

Fixes: #1229

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,7 @@
     "expect": "^1.9.0",
     "expect-jsx": "^3.0.0",
     "express": "^4.13.3",
-    "formatjs-extract-cldr-data": "^4.0.0",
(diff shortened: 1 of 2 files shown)


essensx:
Shouldn't the `yarn.lock` file with updated dependency be included?

okuryu:
@essensx good catch! I bumped `yarn.lock` file.

tranvansang:
How about pull request #1214
Currently it is blocking in production

PR #1236, opened 2019-01-30
[fix] Compatibility of injectIntl and static contextType v2

NicoVega:
Redone the last PR since the Original contributor didnt do the CLA stuff.

React 16.6 introduces a new static contextType. If you use it in a component wrapped by injectIntl and old hoist-non-react-statics it will copy contextType to the wrapper component thus breaking context integration. Old versions were unaware of this so they would copy them through. React will then look at contextType and ignore contextTypes that injectIntl uses. It will then complain it can't find the context provided by react-intl.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-intl",
-  "version": "2.8.0",
+  "version": "2.9.0",
   "description": "Internationalize React apps. This library provides React components and an API to format dates, numbers, and strings, including pluralization and handling translations.",
   "keywords": [
     "intl",
@@ -64,7 +64,7 @@
     }
   },
   "dependencies": {
-    "hoist-non-react-statics": "^2.5.5",
+    "hoist-non-react-statics": "3.3.0",
     "intl-format-cache": "^2.0.5",
     "intl-messageformat": "^2.1.0",
     "intl-relativeformat": "^2.1.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-intl",
-  "version": "2.9.0",
+  "version": "2.8.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -64,7 +64,7 @@
     }
   },
   "dependencies": {
-    "hoist-non-react-statics": "^2.5.5",
+    "hoist-non-react-statics": "3.3.0",
```
joxxoxo: how about "^3.3.0"?

joxxoxo:
Need this to fix `[React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.` for component with `injectIntl` + `contextType`

fletcher91:
For @joxxoxo and others, the dependency restriction can be [circumvented with yarn](https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-flat) via a `resolutions` block in `package.json`;

```
"resolutions": {
    "hoist-non-react-statics": "3.3.0"
  },
```

kodedninja:
Will there be a release with this soon(-ish)? It's pretty critical with React 16.

papasmile:
Scope is going to have to be a bit larger since Hoist 3.x [1] drops support for React 14.x, but this project still supports it[2]

[1] https://github.com/mridgway/hoist-non-react-statics/blob/master/CHANGELOG.md#300-july-27-2018
[2] https://github.com/yahoo/react-intl/blob/v2.8.0/package.json#L75

papasmile:
Wait is that a duplicate of https://github.com/yahoo/react-intl/pull/1221 ?

redonkulus:
Closing in favor of https://github.com/formatjs/react-intl/pull/1272

PR #1245, opened 2019-02-18
injectIntl displayName

Qeneke:
Better displayName "Unknown" in production environment


StorytellerCZ:
Is this really necessary? If anything I would say that would make things difficult to pin down an issue in production.

papasmile:
@Qeneke can you maybe explain what you mean by 'Better' display name?  E.g. what problem are you trying to solve?

redonkulus:
Closing, please re-open if needed.

PR #1272, opened 2019-05-09
#1269 - contextTypes overriding

### LINKED ISSUE #1269: contextType of injectIntl is overridden by child component contextType
### Expected behavior
`contextType` of react-intl should not be overridden

### Current behavior
`contextType` of react-intl is overridden by child component's `contextType` field. The reason is that: react-intl uses an old version (2.5.5) of `hoist-non-react-statics` which doesn't include the new `contextType` field in the statics blacklist.

Updating the dependency to the latest version should fix the problem!

1. Add own context to a react component using `Component.contextType = Context` or inside a class with babel `static contextType = Context;`
2. Wrapp `injectIntl` around the component

See live example: https://codesandbox.io/s/yv1q6xq66x

### Your Environment

| Executable | Version |
| ---: | :--- |
| `npm ls react-intl` | 2.8.0 |
| `npm ls react` | 16.8.6 |

| OS | Version |
| --- | --- |
| Arch Linux | 5.0.9 |

| Browser | Version |
| --- | --- |
| Chrome | 73.0.3683.103 |

tybenz:
Upgrades a dep to avoid the context override issue in #1269


yahoocla:

tybenz:
@DragonRaider5 @ericf ... Can we get a maintainer to look at this PR? This is a blocking issue for my team & a really simple change.

redonkulus:
@tybenz are you using version 3 now or do you need this for v2 branch?

papasmile:
We have three of that same PR.... I think #1221 was first but has a conflict.

Also package-lock.json was removed in 3.0.0-beta but shouldn't it be added to `.gitignore`?

redonkulus:
@papasmile I think the package-lock.json should be added once we move back to npm.

tybenz:
@papasmile need it for v2. should i put up a separate PR?

redonkulus:
@tybenz Yup, keep this one open, as we can upgrade v3 as well. You can open another PR against the [2.x branch](https://github.com/formatjs/react-intl/tree/2.x).

tybenz:
@redonkulus ok updated this PR and opened one for 2.x

redonkulus:
Released in v3.0.0-beta-2

PR #1273, opened 2019-05-09
Remove deprecated lifecycles in FormattedRelative

janicduplessis:
This removes deprecated react 16 lifecycles.

- `componentWillUpdate` only performs a side effect (`setTimeout`) so it is fine to move it to `componentDidUpdate`.
- `componentWillReceiveProps` is replaced with `getDerivedStateFromProps`. The main change required is to save the previous value in state so we can check if it changed in `getDerivedStateFromProps`.

### DIFF
--- a/src/components/relative.js
+++ b/src/components/relative.js
@@ -131,11 +131,11 @@ class FormattedRelative extends Component {
     this.scheduleNextUpdate(this.props, this.state);
   }
 
-  static getDerivedStateFromProps({ value }, { prevValue }) {
+  static getDerivedStateFromProps({ value, intl }, { prevValue }) {
     // When the `props.value` date changes, `state.now` needs to be updated,
     // and the next update can be rescheduled.
     if (!isSameDate(value, prevValue)) {
-      return {now: this.props.intl.now(), prevValue: value};
+      return {now: intl.now(), prevValue: value};
     }
     return null;
   }
--- a/src/components/relative.js
+++ b/src/components/relative.js
@@ -92,7 +92,7 @@ class FormattedRelative extends Component {
 
     // `now` is stored as state so that `render()` remains a function of
     // props + state, instead of accessing `Date.now()` inside `render()`.
-    this.state = {now};
+    this.state = {now, prevValue: props.value};
   }
 
   scheduleNextUpdate(props, state) {
@@ -131,20 +131,21 @@ class FormattedRelative extends Component {
     this.scheduleNextUpdate(this.props, this.state);
   }
 
-  componentWillReceiveProps({value: nextValue}) {
+  static getDerivedStateFromProps({ value }, { prevValue }) {
     // When the `props.value` date changes, `state.now` needs to be updated,
     // and the next update can be rescheduled.
-    if (!isSameDate(nextValue, this.props.value)) {
-      this.setState({now: this.props.intl.now()});
+    if (!isSameDate(value, prevValue)) {
+      return {now: this.props.intl.now(), prevValue: value};
     }
+    return null;
   }
 
   shouldComponentUpdate(...next) {
     return shouldIntlComponentUpdate(this, ...next);
   }
 
-  componentWillUpdate(nextProps, nextState) {
-    this.scheduleNextUpdate(nextProps, nextState);
+  componentDidUpdate() {
+    this.scheduleNextUpdate(this.props, this.state);
   }
 
   componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


yahoocla:

DragonRaider5:
Hey @janicduplessis,

thanks for your support!
Have you had a look at `react-intl@next`? It drops support for older react versions and by doing so allowed us to upgrade to the new context API & lifecycle hooks.

I hope this fixes your issues/warnings and the changes you did didn't cost you too much time.

Best regards

janicduplessis:
@DragonRaider5 The changes are based on the master branch of this repo, I'm pretty sure they are still needed unless I missed something.

DragonRaider5:
@janicduplessis yes my bad, you seem to be right - I am sorry. This repo was recently moved and I didn't notice my PR got merged (yesterday or so).

Seems like I've missed this component while updating, my bad again 🤣
Have you already signed the CLA, so we can merge this PR when review finished? 👍 

@redonkulus can you trigger this PR to be rebuilt by Travis?

janicduplessis:
@DragonRaider5 awesome! Yes I did sign the cla but it seems like the bot did not update the status

redonkulus:
@DragonRaider5 I restarted the build.
@janicduplessis unfortunately our CLA bot doesn't update the status after you have signed it.

DragonRaider5:
Thanks @redonkulus for the restart.

How do we handle the CLA? I think another push to the repo is needed in order to get the bot to check it again. Am I right?

redonkulus:
@janicduplessis looks like some of the tests failed due to `this.props` being undefined. Can you take a look?

src/components/relative.js
```diff
@@ -131,20 +131,21 @@ class FormattedRelative extends Component {
...
   }
 
-  componentWillReceiveProps({value: nextValue}) {
+  static getDerivedStateFromProps({ value }, { prevValue }) {
     // When the `props.value` date changes, `state.now` needs to be updated,
     // and the next update can be rescheduled.
-    if (!isSameDate(nextValue, this.props.value)) {
-      this.setState({now: this.props.intl.now()});
+    if (!isSameDate(value, prevValue)) {
+      return {now: this.props.intl.now(), prevValue: value};
```
DragonRaider5: The issue is located here, `this.props.intl.now()` needs to be changed to `intl.now()` and `intl` must be deconstructed from the current props being passed to `static getDerivedStateFromProps()`, the same way as `value` is.

DragonRaider5:
Hahaha race condition :rofl:

janicduplessis:
Oups looks like I didn’t test the getDerivedStateFromProps case properly, thanks tests.

Looks like some are still failing, probably related to some small timing difference with the new react lifecycles, I’ll look at it later today.

DragonRaider5:
The failing test is caused by `static getDerivedStateFromProps()`, as the test for infinite values given as the `value` prop expects `setState` to have been called exactly 1 time - however this now is handled by `static getDerivedStateFromProps()`, so this requires changing.

You can also test locally, meaning you won't need to push every time :D

janicduplessis:
@DragonRaider5 Alright, the tests are fixed!

DragonRaider5:
@janicduplessis thanks a lot!
However this doesn't fix the test in a meaningful way. The spy on `setState` needs to be replaced with a spy on `static getDerivedStateFromProps` to make sure the thing which was intended to be tested really works.

The test you changed was about preventing very rapid update loops in case bad values for the `value` prop where given.
Previously the time was being updated using `setState` which is why the amount of calls to `setState` was being tested. However this now is dealt with by `static getDerivedStateFromProps` which is why this now has to be spied on.

test/unit/components/relative.js
```diff
@@ -70,7 +70,7 @@ describe('<FormattedRelative>', () => {
 
         // Should avoid update scheduling tight-loop.
         await sleep(10);
-        expect(setState.calls.length).toBe(1, '`setState()` called unexpectedly');
```
DragonRaider5: Needs to be changed to spy `static getDerivedStateFromProps`.

DragonRaider5: I think this would work (replace `setState` and `spySetState`):
```js
const spyGetDerivedStateFromProps = () => {
  return spyOn(
    require('../../../src/components/relative').BaseFormattedRelative.constructor,
    'getDerivedStateFromProps'
  );
}
```
⚠️ not tested

janicduplessis:
@DragonRaider5 Updated the test!

DragonRaider5:
@janicduplessis beatutiful *-* thank you very much!

@redonkulus I'll merge, can you pls push to npm?

redonkulus:
@janicduplessis Thank you for your contributions.

@DragonRaider5 I will merge and release.

redonkulus:
Released in v3.0.0-beta-3

PR #1277, opened 2019-05-15
Upgrading hoist-non-react-static

tybenz:
This one is for v2

Upgrades a dep to avoid the context override issue in #1269

Goes with https://github.com/formatjs/react-intl/pull/1272

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     }
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


redonkulus:
Thanks!

redonkulus:
Released in v2.9.0

Issue #1274, opened 2019-05-14
babel config is wrong in the examples


keremciu:
It looks the config is for 'react-intl' babel-plugin but it's not there

https://github.com/formatjs/react-intl/blob/4109ef13e8a7b3b2af5f8d5fead06af9e5cfbb95/examples/translations/.babelrc#L7

redonkulus:
@keremciu have you tried running the example locally? if so, does it fail?

keremciu:
@redonkulus yes but the problem is this config for this babel plugin doesn't make sense.

as you can see below link there's no messagesDir config inside this babel plugin.
https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/blob/master/src/index.js 

you can see the problem in this commit, `messagesDir` was a config parameter for react-intl plugin.
https://github.com/formatjs/react-intl/commit/fcf9d4fc800efde553044cbdd27abc3cf7625d55#diff-12ea091f62ce6c62baeea4826fc5f9f8

redonkulus:
@keremciu do you mind submitting a PR to fix this?

keremciu:
@redonkulus yep yep, I created one https://github.com/formatjs/react-intl/pull/1278

redonkulus:
Merged! Thanks for helping.

PR #1278, opened 2019-05-16
fix babel config problem in examples #1274

keremciu:
There was a mistake inside .babelrc files about plugins part. I added `babel-plugin-react-intl` plugin inside .babelrc file as mentioned inside README.md file.

I tested examples they're working but the only case is there, we need to add dependencies manually maybe we could add them into package.json?


yahoocla:

PR #1279, opened 2019-05-16
feat: add stale bot to repository

redonkulus:
As of this issue creation, this repository has ~189 open issues, many of which are up to 4 years old. Most of these issues will never be worked on. Having all these issues open makes it hard to prioritize them. I would prefer to slowly close stale and outdated issues rather then let them sit open with no intention to work on them. 

This PR introduces [Stale bot](https://github.com/probot/stale) to help us manage this automatically. Similar models been leveraged successfully in other successful OS projects like webpack. 

Stale bot also answers the question [`Is closing stale issues really a good idea?`](https://github.com/probot/stale#is-closing-stale-issues-really-a-good-idea) fairly well too:

> In an ideal world with infinite resources, there would be no need for this app.

> But in any successful software project, there's always more work to do than people to do it. As more and more work piles up, it becomes paralyzing. Just making decisions about what work should and shouldn't get done can exhaust all available resources. In the experience of the maintainers of this app—and the hundreds of other projects and organizations that use it—focusing on issues that are actively affecting humans is an effective method for prioritizing work.

> To some, a robot trying to close stale issues may seem inhospitable or offensive to contributors. But the alternative is to disrespect them by setting false expectations and implicitly ignoring their work. This app makes it explicit: if work is not progressing, then it's stale. A comment is all it takes to keep the conversation alive.

If anyone has a strong reason to not do this, I'm open to hearing them out. However, I believe this will help the project out in the long run.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,62 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Only issues or pull requests with all of these labels are check if stale. Defaults to `[]` (disabled)
+onlyLabels: []
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+  - security
+  - "[Status] Maybe Later"
+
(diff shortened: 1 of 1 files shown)


DragonRaider5:
In my opinion this is a very nice idea and a good enhancement. I thought about going through them one by one, but this would waste a lot of time as many of those issues simply are out of date or use-case specific, meaning that if the user don't care anymore, they've become pointless.

DragonRaider5:
However, as a side note, I think we need to further encourage people to fix what they report - as this is the concept of OpenSource.
I also think that we did a much better job at this in recent history, just wanted to give a lil heads-up.

I love the activity :heart:

juandopazo:
👍

Issue #1224, opened 2018-12-19
[Question] Any plan to use new react hooks?


Betree:
So React introduced [hooks](https://reactjs.org/docs/hooks-intro.html).

React-i18next have [implemented](https://react.i18next.com/experimental/using-with-hooks) it as a proof of concept, and their API can now be used like that:

```es6
import React, { Suspense } from 'react';
import { useTranslation } from 'react-i18next/hooks';
​
function MyComponent() {
  const [t, i18n] = useTranslation();
​
  return <h1>{t('Welcome to React')}</h1>
}
```

The simplicity of this code is very attractive. Can we imagine something similar for `react-intl`?

Boshen:
To workaround, you can easily take the old intl context and turn it into hooks:
```typescript
import * as React from 'react'
import { useContext } from 'react'
import { injectIntl, InjectedIntl } from 'react-intl'

export const IntlContext = React.createContext<InjectedIntl>({} as InjectedIntl)

// turn the old context into the new context
export const InjectIntlContext = injectIntl(({ intl, children }) => {
  return (
    <IntlContext.Provider value={intl}>
      { children }
    </IntlContext.Provider>
  )
})

// the format message hook
export const useFormatMessage = () => {
  const intl = useContext(IntlContext)
  return intl.formatMessage
}
```

then use it something like:
```typescript
const TestIntl = () => {
  const t = useFormatMessage()
  return (
    <div>
      { t({ id: 'preferences' }) }
    </div>
  )
}

const App = () => 
  <IntlProvider locale={intlLocale} key={intlLocale} messages={messages} defaultLocale='en-gb'>
    <InjectIntlContext>
      <TestIntl />
    </InjectIntlContext>
  </IntlProvider>
```

DragonRaider5:
With the new version (which utilizes the new context API #1186) this would be quiet easy to do I guess.

leoyli:
@Boshen have demonstrate how easy it would be and it looks have very little impact on bundling size. I personally really want to get rid of the `injextIntl` HOC to make my tree flatter and easier for typing my component interface in TS, @redonkulus, @DragonRaider5 would you guys consider to add `useIntl` hook here?

redonkulus:
@leoyli I think that’s a good idea. Would you like to take a stab at it and help contribute?

leoyli:
@redonkulus, sure.

Here is what I would proposed: `useIntl` hook to mimic `injectIntl`.

[...]

Issue #1281, opened 2019-05-24
Update Migration Guide


redonkulus:
We need to update the [v3.0.0 Migration guide](https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md) to ensure it is easy for developers to upgrade.

We can follow the same format as previous releases.

DragonRaider5:
WIP :smile:

leoyli:
#1282, some documentation needed for how to use this (it's is a simple replacement for `injectIntl` HOC for function components).

redonkulus:
@DragonRaider5 the upgrade doc has been moved to https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md. 

@leoyli good suggestion. Want to PR that to the doc above?

leoyli:
@redonkulus, `Components` also have no `injectIntl` documented, maybe it belongs to other docs? (haven't look into details)

redonkulus:
@leoyli Its actually in the [API](https://github.com/formatjs/react-intl/blob/master/docs/API.md#injection-api) docs. We should add `useIntl` to that doc as well.

redonkulus:
@leoyli do you want to open a PR for adding `useIntl` to the API docs?

@DragonRaider5 @longlho anything else we need to call out for v3 migration?

leoyli:
@redonkulus, I can do it in the weekend, if that is okay.

redonkulus:
Its no problem, if there are no other tasks or bugs opened for v3, we can probably release it next week.

leoyli:
@longlho, Could you elaborated a bit more on "removing `addLocaleData`" part in the upgrade guide for me?! I found I get a bit confused on what should I do if I have `addLocaleData` used.  Am I just simply remove it? And if I have to support IE11, then should I just keep it but adding polyfills in my build entries instead?! 

If you have good answer for me, please consider adding these into the doc.  I believe it would be super helpful to all users :pray:

longlho:
Yeah for modern browsers where `Intl.RelativeTimeFormat` is supported, you don't need to do anything. For IE11 the locale data would be part of the `Intl.RelativeTimeFormat` polyfill here: https://github.com/formatjs/formatjs/tree/master/packages/intl-relativetimeformat

I'll add more details to the doc

leoyli:
[...]

PR #1282, opened 2019-05-26
#1224 - Add useIntl hook

leoyli:
**Disclaimer**: I did not sign Yahoo!'s CLA's since this project is no longer supported by Yahoo!.

Follow up our discussions in #1224; it would be nice to have `useIntl` as a shortcut to consume `intl` object directly in a function component. This PR provide this feature.

## Usage

```js
import { useIntl } from 'react-intl';
```

## Tests Coverage

```
--------------------------|----------|----------|----------|----------|-------------------|
File                      |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
--------------------------|----------|----------|----------|----------|-------------------|
All files                 |    97.73 |    92.34 |      100 |    97.71 |                   |
 src                      |    98.09 |    91.13 |      100 |    98.06 |                   |
  define-messages.js      |      100 |      100 |      100 |      100 |                   |
  format.js               |    97.85 |    91.46 |      100 |    97.85 |           171,174 |
  index.js                |      100 |      100 |      100 |      100 |                   |
  locale-data-registry.js |      100 |    93.33 |      100 |      100 |                14 |
  plural.js               |      100 |      100 |      100 |      100 |                   |
  react-intl.js           |      100 |      100 |      100 |      100 |                   |
  types.js                |      100 |      100 |      100 |      100 |                   |
  utils.js                |    96.55 |    88.46 |      100 |    96.43 |                66 |
 src/components           |    97.45 |    93.88 |      100 |    97.44 |                   |
  date.js                 |      100 |      100 |      100 |      100 |                   |
  html-message.js         |      100 |      100 |      100 |      100 |                   |
  message.js              |      100 |     96.3 |      100 |      100 |                21 |
  number.js               |      100 |      100 |      100 |      100 |                   |
  plural.js               |      100 |      100 |      100 |      100 |                   |
  provider.js             |      100 |      100 |      100 |      100 |                   |
  relative.js             |    90.57 |    83.33 |      100 |    90.57 |    33,34,39,49,51 |
  time.js                 |      100 |      100 |      100 |      100 |                   |
  useIntl.js              |      100 |      100 |      100 |      100 |                   |
  withIntl.js             |      100 |      100 |      100 |      100 |                   |
--------------------------|----------|----------|----------|----------|-------------------|

=============================== Coverage summary ===============================
Statements   : 97.73% ( 345/353 )
Branches     : 92.34% ( 205/222 )
Functions    : 100% ( 79/79 )
Lines        : 97.71% ( 342/350 )
================================================================================
Test Suites: 18 passed, 18 total
Tests:       275 passed, 275 total
Snapshots:   0 total
Time:        15.26s

```


test/unit/components/useIntl.js
```diff
@@ -0,0 +1,35 @@
+import expect, {spyOn, createSpy} from 'expect';
```
DragonRaider5: Please remove the `expect` import in order to use the jest default.

src/react-intl.js
```diff
@@ -14,6 +14,7 @@ export {intlShape} from './types';
 export {default as defineMessages} from './define-messages';
 
 export {default as injectIntl, Context as IntlContext} from './components/withIntl';
+export {default as useIntl} from './components/useIntl';
```
DragonRaider5: Please add a check towards wether or not the React version is gte. `16.8.0` (supports hooks), eg.:

```js
const [ major, minor ] = React.version.split('.')
if (major >= 17 || minor >= 8) {
  // export
}
```

leoyli: I was thinking this too, but does it really required?! The user won't be able to execute hook anyway (since `useContext` is not a function). So I think it's kind of redundant... but I would be happy to add this if there are other concerns (e.g. it allows tree shaking to work or something else?)

DragonRaider5: You're right, it shouldn't make any difference after all :rofl::+1:

test/unit/components/useIntl.js
```diff
@@ -0,0 +1,34 @@
...
+describe('useIntl() hook', () => {
+  it('throws when <IntlProvider> is missing from ancestry', () => {
+    const consoleError = jest.spyOn(console, 'error').mockImplementation(() => {}); // surpress console error from JSDom
+    expect(() => mount(<FunctionComponent />)).toThrow(
+      '[React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry.'
+    );
+    consoleError.mockRestore();
+  });
+
+  it('spy on the returned value from hook', () => {
```
DragonRaider5: We try to make the test names describe what is being tested (and ideally use the it `it` in the sentence), e.g.:
```js
it('hooks onto the intl context')
```
This makes the test status overview a lot easier to read :smile:

leoyli: Nice test naming suggestion! It made a lot sense now 🚀!

redonkulus:
@leoyli @DragonRaider5 should `peerDependencies` be updated since it relies on `useContext`?

DragonRaider5:
@redonkulus no, see the comments on `src/react-intl` for details :+1:

leoyli:
@redonkulus, this hook is just a sugar for those folks who wants to use hook instead of HOC.  They will have to be responsible to get the correct version of React (with the one support hooks). Since the API from react is fresh clean, and the hook is used by lazy evaluation at run-time, we don't need to worry about peerDependencies or rise the minimum version support to 16.8. I think then everyone should be happy, it's a win-win 🎉 .

papasmile:
@leoyli nice stuff, can we get some:
* Code comments on `useIntl.js`
* Documentation in markdown
* Extra credit: an addition to /examples for hook usage..

DragonRaider5:
@papasmile
- I'll be taking care of the documentation if he doesn't
- there is literally nothing to comment on, it's like 10 lines of code and this feature is purely for convenience
- an example should not be needed, again because this is only convenience. The hook works like every hook works and is simply an alternative to using the HOC for injection of the intl context

papasmile:
At a minimum some English makes things both known and easier to use.  I'd suggest:

* A description on top of  L5 on`useIntl.js`
* And an addition to Wiki under or around present injection API: https://github.com/formatjs/react-intl/wiki/API#injection-api

src/components/useIntl.js
```diff
@@ -0,0 +1,9 @@
+import {useContext} from 'react';
+import {IntlContext} from './../react-intl';
+import {invariantIntlContext} from '../utils';
+
+export default function useIntl() {
+  const intl = useContext(IntlContext);
+  invariantIntlContext({ intl });
+  return intl;
```
DragonRaider5: As I suggested in the issue related to this PR, what do you guys think about changing this to
```js
return [intl.formatMessage, intl];

// Allows for usage as following
const [ t ] = useIntl()

// Currently
const { formatMessage: t } = useIntl()
```

I think this might allow for slightly less code, as 90% of the time people use `formatMessage`.
However it might take away some simplicity.  I think it'd be worth it.

leoyli: I personally think it is beautiful, I'm totally okay with this. As for the simplicity, I think this is also cleaner and idiomatic (think about other lib and the primitive hooks' API), but it surly need more documentation.

Are we agree it is the convenience we want?! I think it's straightforward.

DragonRaider5: I agree. I think it's a good way to go.

leoyli:
@DragonRaider5, @papasmile,

I would be happy to carry on the documentation too, but I think it's pretty straightforward and since @DragonRaider5 is working on #1281, should we separate the concerns here? We can discuss and review the documentation on top of that future PR. 😄

src/components/useIntl.js
leoyli: @DragonRaider5, I've updated my PR, please have a look 😸!

test/unit/components/useIntl.js
```diff
@@ -25,10 +25,9 @@ describe('useIntl() hook', () => {
         <FunctionComponent spy={spy} />
       </IntlProvider>
     );
-    const intlProvider = rendered.find(IntlProvider).childAt(0);
 
-    expect(spy).toHaveBeenCalledWith(
-      intlProvider.instance().getContext()
-    );
+    const providedIntl = rendered.find(IntlProvider).childAt(0).instance().getContext();
+    const expectedHookReturns = [providedIntl.formatMessage, providedIntl];
```
DragonRaider5: You might as well change the name from `providedIntl` to plain `intl` and inline the expect value, to make the code easier to understand and reflect the tested code in an easier to grasp way.
```js
const intl = rendered.find(IntlProvider).childAt(0).instance().getContext();
expect(spy).toHaveBeenCalledWith([ intl.formatMessage, intl ]);
```

leoyli: @DragonRaider5, Nice! All green and look good to me now =)

leoyli:
Add cross reference: DefinitelyTyped/DefinitelyTyped#35812

redonkulus:
@leoyli as an FYI, when packaging and publishing this, I did get some circular dependency warnings from yarn:

```

**Comment by DragonRaider5** [comment]
Oh well that's mb, didn't see that.
@redonkulus want me to fix?

**Comment by redonkulus** [comment]
@DragonRaider5 I’m not sure it will be an issue but if it’s a quick fix, sure let’s do it.

**Comment by StorytellerCZ** [comment]
@DragonRaider5 @redonkulus I pointed out the circular dependency above, with a suggested fix.

**Comment by redonkulus** [comment]
@StorytellerCZ I'm not seeing it, what was it?

**Comment by cknitt** [comment]
I think `useIntl` returning `[formatMessage, intl]` is not a good idea because:

1.) Principle of least surprise: As a user of the API I would expect:

* `useIntl` to return an `intl`
* `useFormatMessage` to return a `formatMessage` function.

`useFormatMessage` is trivial to define as a custom hook (either as part of react-intl or by the user) based on a `useIntl` hook returning an `intl`:

```
export function useFormatMessage() {
  return useIntl().formatMessage;
}
```

(Same for `useFormatDate` etc. - the user can just define the custom hooks he needs.)

2.) Not only will users make errors because they expect `useIntl` to return `intl` - even if they know that a tuple is returned, they will also make errors because they will confuse the order (`[intl, formatMessage]` instead of `[formatMessage, intl]`.

**Comment by alexfedoseev** [comment]
+1 to @cknitt 

It's surprising (i.e. confusing) that `useIntl` returns tuple instead of `intl` itself. I would delegate such controversial api optimizations to user land or at least exposed low-level hook that returns `intl` itself thus it's possible to bind to it and build application level abstractions without jumping through such hoops.

**Comment by jono-allen** [comment]
+1 as well. We tried out the new hook yesterday and works well but we did agree the api was unusual. I would agree to move to an api where it just returns intl instead would be preferable.  Our case, every time we used it we had the following.  `const [_ , intl] = useIntl()`

**Comment by DragonRaider5** [comment]
Maybe the naming should've changed with the return.
Raw (user) hooks are still possible, because the whole context is being exported (as `IntlContext`).

However I think order is not a problem if one knows it's an array.

**Comment by DragonRaider5** [comment]
`IntlContext` being exported to me also is a reason  not to export the basic hook.
However I don't mind changing it, I just wonder why the interest in this is as big? Primary usage is still supposed to be by using the components?

**Comment by alexfedoseev** [comment]
In `bs-react-intl`, we have `useIntl` implementation returning `intl` for quite awhile (a number of codebases already rely on it) and we would like to keep these bindings zero-cost + 100% matching `react-intl` api.

**Comment by cknitt** [comment]
Generally, in React Native you need an intl instance quite often to localize strings for APIs like https://facebook.github.io/react-native/docs/alert, local notifications etc.

**Comment by DragonRaider5** [comment]
@alexfedoseev why would it be *our* concern to match *your* API? I'm confused...
Also if there already was a rebuild of `react-intl` using the new API, why didn't you share it here and why did I have to do that again?

**Comment by DragonRaider5** [comment]
@cknitt is there any problem with the proposed API for doing this?

**Comment by DragonRaider5** [comment]
@redonkulus what do you think regarding the feedback?

**Comment by cknitt** [comment]
@DragonRaider5 This was just to answer your question

>  I just wonder why the interest in this is as big?

Alert/local notifications in React Native are one use case where the intl instance is needed and the FormattedMessage components are not sufficient.

Of course this use case can be supported with the proposed API, too. It's just that the proposed API is unnecessarily confusing and redundant IMHO.

**Comment by alexfedoseev** [comment]
@DragonRaider5 This is not _my_ api. `bs-react-intl` is BuckleScript bindings to the `react-intl` library. So the point is to avoid different apis here. I don’t understand what you mean by your second sentence.

**Comment by fhammerschmidt** [comment]
I agree with @alexfedoseev and @cknitt here. I have multiple occurrences of a custom `useIntl` which returns `intl` only and use it wherever I need strings directly (and no <span/> or <Text/>). I think it is very weird to return both the "main" `intl` object and a sub object `formatMessage` at the same level.

I would keep to userspace what should be in userspace.

**Comment by redonkulus** [comment]
@DragonRaider5 I don't really have a preference either way. If the community agrees that its unnecessary then I'm fine with just keeping it simple and returning `intl`. If users really need `formatMessage` they can always get it from there. I questioned the convenience of it originally, but didn't speak up because it wasn't that big of a deal for me. 

One thing that is interesting is that `useState` also returns a tuple, but I suppose that is more justified as it returns a function that allows you to modify the state later on. Just mentioning this as perhaps maybe this was one reason this format was chosen.

**Comment by cknitt** [comment]
Right, `useState` and `useReducer` do return tuples, but that is for a very good reason (getter and setter/dispatch function). Not every hook needs to return a tuple though, the other "standard" hooks (`useContext`, `useRef`, `useEffect`, `useCallback`, `useMemo`, ...) don't.

Would you like me to create a PR to revert https://github.com/formatjs/react-intl/pull/1282/commits/736da313912e4f81221269e4be865bbb8970e377?

**Comment by DragonRaider5** [comment]
@alexfedoseev thanks for the clarification. In my second sentence I was talking about the new context API as to why `bs-react-intl` already had an API called `useIntl` because this implies `bs-react-intl` had implemented the new context API before us.

@redonkulus this was not the reasoning. The reason was that in the original issue #1224 the question was for an API like the one of `react-i18next`. In our opinion their syntax (without knowing too much of their code) seemed very reasonable, as 90% of our APIs usage is `formatMessage` AFAIK.

@fhammerschmidt @cknitt when you need strings, which functions do you use and in which ratio?

**Comment by cknitt** [comment]
I agree, 90% of the time we will need `formatMessage`, not `formatDate`, `formatNumber` etc.

For such cases, a `useFormatMessage` hook might be helpful if one does not want to destructure `intl` or write `intl.formatMessage(...)`.

For cases where multiple functions are used, destructuring makes most sense:

```js
const { formatMessage, formatDate, formatNumber } = useIntl();
```

leoyli:
I think we have nice feedback here, and all of these are relatively trivial to be implemented.

As @DragonRaider5 mentioned, the original goal is to have `react-i18next` like API, but it seems doesn't resonate quite well within the community here, especially for those people using `react-native`.

And consider the fact that `useIntl` hook v.s. `withIntl` HOC, it would be reasonable to align with their API surface and reduce the need to document their difference exhaustively.

Would be also curious do we really need to have `useFormatMessage` hook? If so, it seems reasonable for people to argue why not add `useFormatDate` too.
---

@DragonRaider5, and sorry for the circular dependencies issue I didn't find out, so just to be clear, you are going to fix the issue here? Or I would be happy to submit my PR (with the above revert) again 🚀 .

DragonRaider5:
@cknitt if so, don't you think writing
```js
const [ , intl ] = useIntl()
```
instead of
```js
const intl = useIntl()
```
for those 10% where we need something other than `formatMessage` is well worth it, as it makes it possible to write
```js
const [ t ] = useIntl()
```
instead of
```js
const { formattedMessage } = useIntl()
```
for the remaining 90%? In addition it allows users to rename the `formatMessage` function with less writing i.o. to make it integrate into code more seemlessly (as seen in the `react-i18next` example, I actually like their thought).

Thanks for your feedback!

@leoyli I can so after we've made up our minds regarding this conversation :+1:

cknitt:
Well, I would prefer the even clearer

```
const t = useFormatMessage();
```

for that main use case. `useIntl` returns `intl`. `useFormatMessage` returns `formatMessage`. Simple. 😉

leoyli:
@cknitt,

I don't think this hook is really needed:

> We were discussed to provide a shortcut as `useFormatMessage` hook since it's 90% use case, I personally okay with that since it looks neat and convenient. But as I said, I don't want to have that implemented and then people start to argue about why not `useFormatXXX` too... So maybe these shortcuts should really do in the user's land?! They can just wrap our `useIntl` hook to do whatever they want. 🤔

---

@DragonRaider5,

I like the approach too, but it seems not that intuitive, for example, what @cknitt said can lead something like this:

```js
const [,{ formatMessage, formatDate, formatNumber }] = useIntl();
```

We do doubly destructure just to get things out 👀, not sure it is good or bad.  But I guess the confusion (or surprise) that people rise perhaps is due to we haven't document the API?

cknitt:
To clarify, I don't think `useFormatMessage` necessarily needs to be in `react-intl`, I'd be absolutely happy to define it in userland based on `useIntl` if/when I need it.

Possibly a few words could be added to the docs that users can define custom hooks for shortcuts like this.

alexfedoseev:
> @alexfedoseev thanks for the clarification. In my second sentence I was talking about the new context API as to why bs-react-intl already had an API called useIntl because this implies bs-react-intl had implemented the new context API before us.

@DragonRaider5 Ah, no, you got me wrong. We never had such implementation, only bindings. At first, we used [this fork](https://www.npmjs.com/package/@4c/react-intl) and after that we used [exported context](https://github.com/alexfedoseev/bs-react-intl/blob/hooks/src/ReactIntl.re#L137-L140) to expose `useIntl`.

redonkulus:
@DragonRaider5 @leoyli I'd say just do what we had originally, `const intl = useIntl()` and let users decide on how they want to expose `formatMessage` in their own apps. Users that want `t` or whatever can just create their own wrapper and re-use that across their apps. Also keeps our API simple and removes questions about supporting other `format*` APIs.

leoyli:
So does here we can draw a line for conclusion? (@DragonRaider5, @redonkulus)

Here is two directions:

1. `useIntl` hook returns a tuple `[formatMessage, intl]`:

This mimic the behaviour of another popular i18n library, `react-i18next`. The reason behind is 90% of time we are using `intl.formatMessage` function to translate a string in a component. Using tuple allow us to expose the function without explicitly naming it, then we can just call it `t` or whatever we like, this is similar to how `useState`, `useReducer` do. But, as people argued, the return value is not intuitive at the first place without dig into documentations.

2. `useIntl` hook returns `intl`:

This seems much intuitive and reflects what is going to be returned from the function.  The down side is we can not take the advantage as multi return of a function, and need to write a bit verbose code (but might be a bit more explicit). Another point in my mind is, if we take `useIntl` to refract `withIntl` HOC, then this makes refractor a breeze (assuming the function get called as `intl.formatMessage(...)`). As a side note, another i18n library `lingui` take this direction.

---

To be honest, after thinking this deeply, I would prefer (2.) since it avoid someone use weird naming that make team working and maintainability harder in a long term. Of course they can go crazy renaming from destructure reassignment, but we setup a barrier to do it, not lower it. It's more an art about API design, and I think we'd better think it twice before hitting a stable v3.

redonkulus:
Thanks everyone for the constructive feedback. I've merged #1295 and published to react-intl@3.0.0-beta-6. In the future, if more hooks are warranted, we can always enhance the API to include them.

PR #1312, opened 2019-06-10
1281: add useIntl hook in API doc

leoyli:
## What I did

- Add `useIntl` API doc:
- Slightly update `injectIntl` example as it is much useful for class components.

Relates #1281 .


redonkulus:
Looks like the build failed because the doc was not formatted by prettier.

leoyli:
Should be good now :)

PR #1516, opened 2019-11-13
chore: Fix Oxford Comma in FormattedList & FormatList examples

tinder-kyleboss:
The docs for Component & API do not accurately convey the output of `FormattedList` and `FormatList`.

Following the [ListFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ListFormat) spec, `FormattedList` & `FormatList` use the [Oxford Comma](https://en.wikipedia.org/wiki/Serial_comma). 

## Example
`<FormattedList type="conjunction" value={['Me', 'myself', 'I']} />`

**Before**
`Me, myself and I`

**After**
`Me, myself, and I`


longlho:
Thanks for the PR. Let me look into it. The complication is that not all en
locales use oxford comma so the ambiguity is the lack of locale in the
example.

longlho:
Yeah so for `en`, it is using [Oxford comma](https://github.com/unicode-cldr/cldr-misc-full/blob/master/main/en/listPatterns.json#L14) but not for [`en-GB`](https://github.com/unicode-cldr/cldr-misc-full/blob/master/main/en-GB/listPatterns.json#L15) so I think there're 2 options here:

1. Leave the example alone, but add a note saying this is for `en-GB`
2. Add the Oxford comma in the examples, and add a note saying this is for `en`

roderickhsiao:
I like option 2, thanks for the details check Long!

tinder-kyleboss:
I like this. I will go ahead and make these changes.

longlho:

PR #1550, opened 2020-01-06
fix: add a parser for chunks emitted from formatHTMLMessage

mallchel:
related to: https://github.com/formatjs/react-intl/issues/1549


longlho:
Thanks a lot for your contributions! Can you add a test also? We have existing mocking of `console.log` that you can mirror to check for warning messages

mallchel:
Yes, I will fix it tomorrow

mallchel:
@longlho We could avoid array in the output of the prepareIntlMessageFormatHtmlOutput.
like this: 
```ts
export const prepareIntlMessageFormatHtmlOutput = (
  chunks: (string | object)[]
): React.ReactElement => React.createElement(React.Fragment, null, ...chunks);
```

But then we will lose information about the number of chunks, although this may be unnecessary

longlho:
hmm if you do that u'll still get key warnings right?

mallchel:
> hmm if you do that u'll still get key warnings right?

No, it will work as in the FormattedMessage component. Without key warnings 
image

longlho:
Ah yeah I forgot. Yup I like that!

mallchel:
@longlho ok, i have changes on the local machine. Will be committed on returning home

longlho:

PR #1659, opened 2020-05-02
fix(@formatjs/cli): add @babel/core as a peer dep

pyrocat101:
This suppresses the warning of unmet peer dependency when installing @formatjs/cli.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "@types/jest": "^25.2.1",
     "@typescript-eslint/eslint-plugin": "^2.29.0",
     "@typescript-eslint/parser": "^2.29.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -30,7 +30,6 @@
     "@types/jest": "^25.2.1",
     "@typescript-eslint/eslint-plugin": "^2.29.0",
     "@typescript-eslint/parser": "^2.29.0",
-    "@yarnpkg/pnpify": "^2.0.0-rc.21",
```
pyrocat101, line 33: Unused

pyrocat101:
Looks like this broke formatjs-website (likely due to hoisting).

PR #1668, opened 2020-05-07
chore: use pretty-quick as pre-commit hook

pyrocat101:
Since it only runs on changed files, it is much faster than `pre-commit`


pyrocat101:
@longlho can you merge this?

Issue #1693, opened 2020-05-28
Need an opt-out option for the XML tag feature


mstarets:
Package: intl-messageformat-parser

Problem description:
The parser violates ICU 4.8 spec by requiring that < and > characters be escaped when they are used in regular message text. Since we are not interested in using the XML tag feature or HTML formatting support, we would like to be able to opt out. 

Proposed solution:
1. Expose 'treatTagsAsText' parser option (exact name TBD)
2. Modify parser.pegjs grammar file to check the option when deciding whether < and > characters require quoting:
function treatTagsAsText() {
     return options && options.treatTagsAsText;
 }
...
unquotedString = $(x:. &{
    return (
        (x !== '<' || treatTagsAsText()) &&
        x !== '{' &&
        !(isInPluralOption() && x === '#') &&
        !(isNestedMessageText() && x === '}') &&
        !(isNestedMessageText() && x === '>' && !treatTagsAsText())
    );
} / '\n')

I'd be happy to submit a pull request if people think that this request is reasonable.

longlho:
You can pin `intl-messageformat-parser` to the `v7`. Dupe of #1648

mstarets:
I just wanted to point out that this issue is different from 1648. I am not proposing to reverse the change you made in 8.0. I am asking to add an opt-out flag that would allow us to not require escaping for < and >. All of your existing customers won't be affected at all. The code changes are minimal, and should be easy to maintain.

Pining to an old version is not an option for us. We would like to get your fixes/updates without having to fork and merge.

longlho:
What's your use case actually? Do you use any of the top level libs like `intl-messageformat` or `react-intl`?

mstarets:
Hey Long, we have our own component framework, and are not using React and react-intl. We are switching our translated message format to ICU 4.8, and decided to use your library (because we liked it a lot!) in two different areas:
1) Build-time transformation of message patterns into formatter functions (along with TypeScript definitions for them) that take message arguments and return fully formatted messages. Here we use only intl-messageformat-parser.
2) Our browser-based designer application is using intl-messageformat (after being being processed with rollup) on the client to format messages shown in the preview pane.
One of our internationalization engineers noticed that the parser is throwing errors for messages like "please enter a number > 5", while ICU does not specify that these characters require escaping with an apostrophe. We figured that the brackets are used to support call-outs to custom formatters to implement message formatting. We have no such requirements though.
[...]

PR #1699, opened 2020-06-03
feat(intl-messageformat-parser): Add 'ignoreTag' option

mstarets:
The option may be used by projects that wish to treat angle bracket characters like regular messages characters that do not require escaping. Setting this option would mean opting out of the tag feature, as the parser will never create a node with type "tag" (8)

See issue #1648 and issue #1693

### DIFF
--- a/packages/intl-messageformat-parser/src/parser.pegjs
+++ b/packages/intl-messageformat-parser/src/parser.pegjs
@@ -25,8 +25,8 @@ See the accompanying LICENSE file for terms.
         }: {}
     }
 
-    function noTagSupport() {
-        return options && options.noTagSupport;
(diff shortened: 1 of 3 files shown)


longlho:
Thanks a lot for your contributions! Would you mind changing the option to `ignoreTag`? Otherwise LGTM

mstarets:
Hey Long,
Thanks. I will change the option to 'ignoreTag'.
Max
    On Wednesday, June 3, 2020, 06:31:51 PM EDT, Long Ho <notifications@github.com> wrote:  
 
 


Thanks a lot for your contributions! Would you mind changing the option to ignoreTag? Otherwise LGTM

PR #1741, opened 2020-06-23
fix(react-intl): fix d.ts rollup

### LINKED ISSUE #1739: Cannot find module './injectIntl' or its corresponding type declarations
**Which package?**

`packages/react-intl`, version `@4.6.10`.

**Describe the bug**

When compiling a TypeScript project that uses `useIntl` and `defineMessages`, the compilation fails as follows:

```sh
node_modules/react-intl/lib/react-intl.d.ts:95:57 - error TS2307: Cannot find module './injectIntl' or its corresponding type declarations.

95 export declare const FormattedPlural: React_2.FC<import("./injectIntl").WithIntlProps<Props_2>> & {
                                                           ~~~~~~~~~~~~~~
```

**To Reproduce**

1. Any TypeScript projects
2. Run `tsc` to compile the project

**Expected behavior**
No error occurs, projects get compiled successfully.

**Desktop (please complete the following information):**

- OS: [Mac OS v10.15.5]
- NodeJS [v12.16.2]
- react-intl Version [4.6.10]

**Additional context**

Version `4.6.9` doesn't have this issue.
After some digging, I'm guessing the issue was introduced by the `4.6.10` release(2 days ago) that adds children typing to `injectIntl`, the generated `react-intl.d.ts` includes an `import('./injectIntl')` statement, but the`./injectIntl` module stays in `./components/` directory.

pyrocat101:
Fixes #1739

Bump the version of api-extractor, which now aborts the build if it sees `import(...)` in the type annotation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "test": "yarn run prettier:check && lerna run test"
   },
   "devDependencies": {
-    "@microsoft/api-extractor": "^7.7.13",
+    "@microsoft/api-extractor": "^7.8.12",
     "@rollup/plugin-json": "^4.0.3",
     "@types/jest": "^25.2.1",
     "@typescript-eslint/eslint-plugin": "^2.29.0",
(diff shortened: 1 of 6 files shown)


pyrocat101:
@longlho master is failing with plural rule tests... can you merge this one?

PR #1872, opened 2020-07-24
Update testing-with-react-intl.md for DOM rendering to use forwardRef

ghost:
withRef has been replaced with fowardRef for testing using DOM rendering (this breaking change was from [v2-v3]( https://formatjs.io/docs/react-intl/upgrade-guide-3x#migrate-withref-to-forwardref)), the docs have the old implementation using withRef, I've updated the example to use forwardRef:


longlho:

longlho:
@SarahLancaster our active branch is `main` so would u mind PR-ing into that instead of `master`?

longlho:
I've merged this into `main` thanks @SarahLancaster !

PR #1926, opened 2020-08-05
feat(@formatjs/ts-transformer): allow to override user defined message id via `overrideIdFn` as function

nodkz:
This PR introduces the following logic:
1) overrideIdFn is a string
  - if msg.id is defined leave it unchanged
  - if msg.id is empty then use interpolateName with provided template
2) overrideIdFn is a function
  - pass msg.id to function as first arg. Allow to user decide what to do with id - it may be empty or contain already defined value

### DIFF
--- a/packages/ts-transformer/src/transform.ts
+++ b/packages/ts-transformer/src/transform.ts
@@ -164,18 +164,20 @@ function extractMessageDescriptor(
   if (!msg.defaultMessage && !msg.id) {
     return;
   }
-  if (!msg.id && msg.defaultMessage && overrideIdFn) {
+  if (msg.defaultMessage && overrideIdFn) {
     switch (typeof overrideIdFn) {
       case 'string':
-        msg.id = interpolateName(
-          {sourcePath: sf.fileName} as any,
-          overrideIdFn,
-          {
-            content: msg.description
-              ? `${msg.defaultMessage}#${msg.description}`
-              : msg.defaultMessage,
-          }
-        );
+        if (!msg.id) {
+          msg.id = interpolateName(
+            {sourcePath: sf.fileName} as any,
+            overrideIdFn,
+            {
+              content: msg.description
+                ? `${msg.defaultMessage}#${msg.description}`
+                : msg.defaultMessage,
+            }
+          );
+        }
         break;
       case 'function':
         msg.id = overrideIdFn(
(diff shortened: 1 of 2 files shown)


nodkz:
I cannot start bazel on my macbook:

Screen Shot 2020-08-05 at 20 49 12

So I launch Jest manually and update existing snapshots without bazel.

nodkz:
I spent several hours to bazel without luck. @longlho please run tests on your machine before merge. Of course, if you accept these changes.

longlho:
looks like you forgot to initialize `git submodule` per https://github.com/formatjs/formatjs#development

longlho:
looks like this change is not backwards compatible. Can you explain the use case for #2 above? Why does `overrideIdFn is a function` need to behave differently from when it's a pattern?

nodkz:
I expect that `ts-transformer`'s and `babel-plugin-react-intl`'s function `overrideIdFn` works in the same manner. BUT currently it's not.

With babel, I can override `msg.id` provided by user. With `ts` transformer I cannot do this. Take a look at the following code – `overrideIdFn` is called only when `msg.id` is empty:
https://github.com/formatjs/formatjs/blob/9e48e2c9d9ac090690b4f88731f99d3258544544/packages/ts-transformer/src/transform.ts#L167

But with babel you can override it even if `msg.id` already defined by user –
https://github.com/formatjs/formatjs/blob/9e48e2c9d9ac090690b4f88731f99d3258544544/packages/babel-plugin-react-intl/index.ts#L202-L214

---

Anyway, it's quite strange that `overrideIdFn` has the following typedef in TypeScript where first arg `id` always is empty (because this function is never called if `msg.id` is present):
https://github.com/formatjs/formatjs/blob/9e48e2c9d9ac090690b4f88731f99d3258544544/packages/ts-transformer/src/transform.ts#L11-L16

So, in other words, I feel that the behavior for `overrideIdFn` should be 
- it's `template string` then use it only if `msg.id` is empty. 
- if it's a function then use it for every message (and no matter exists `msg.id` or not). User can realize any logic what he needs, like in my case: if `msg.id` exists I want to ensure that exists prefix for `msg.id` and if it does not then add it. Like I did in my app with babel plugin. But for a package where is used typescript compilation - I cannot reuse the same logic. I'm not able to modify existed id.

---

PS. About bazel. My bad and you're right - I didn't init submodules.

longlho:

PR #2649, opened 2021-02-26
docs: add example config for Vue-CLI projects

yrd:
This is a follow-up to #2643 which adds some documentation for Vue-CLI projects.


yrd:
Yeah, forget this. I just saw your comment in https://github.com/formatjs/formatjs/issues/2643#issuecomment-785627781. Don't think this needs any more documentation then 😉

PR #2852, opened 2021-04-27
fix(react-intl) update react provider to match GetDerivedStateFromProps declaration.

ghost:
Currently the `IntlProvider` is not assignable to `React.ElementType` as the signatures of `getDerivedStateFromProps` do not match.

This PR makes the props Read-only as per https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v16/index.d.ts#L646

Issue can be seen here https://codesandbox.io/s/recursing-villani-mv4v5?file=/src/App.tsx:234-246


longlho:

PR #2931, opened 2021-05-28
feat(vue-intl): use the same injection key for composition function and plugin

glen-84:
https://github.com/formatjs/formatjs/discussions/2924

### DIFF
--- a/packages/vue-intl/index.ts
+++ b/packages/vue-intl/index.ts
@@ -1,6 +1,7 @@
 import {MessageDescriptor} from '@formatjs/intl'
 export * from './plugin'
 export * from './provider'
+export {intlKey} from './injection-key'
 export {
   IntlShape,
   IntlConfig,
--- a/packages/vue-intl/injection-key.ts
+++ b/packages/vue-intl/injection-key.ts
@@ -1,0 +1,1 @@
+export const intlKey = Symbol()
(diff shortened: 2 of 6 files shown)


glen-84:
@longlho,

The problem with a symbol is that we have to export it to make it available to consumers.

[This](https://github.com/vuejs/vuex/blob/7744e74f50ff41fc7084466728f1a033610f43b5/src/injectKey.js) is what vuex does – they use a string, and also export it, so the consumer can either use the string, or import the key.

Which option do you prefer?

1. Follow vuex, and use a string (`export const intlKey = 'intl'`).
2. Just export a symbol, and the consumer will always need to import it when using `inject` (`inject: {intl: intlKey}`).

longlho:
Let's export the symbol. Thanks a lot for the PR!

glen-84:
This is a breaking change for anyone currently using `inject: ['intl']`. 😞

glen-84:
@longlho,

What is your preference?

1. Release another major version.
2. Switch to option 1 above (the `'intl'` string), which will not be a breaking change.

longlho:
Yeah bump major is fine by me. You don't have to bump manually just need to follow conventional changelog to declare breaking change

glen-84:
> BREAKING CHANGE: The injection key is now a symbol and must be imported from `vue-intl`

Should I force-push an update to the commit message? (sorry about all the questions)

longlho:
Yup

glen-84:
@longlho,

I'm sorry to be a pain, but do you know roughly when this change might be released? 🙂

Thanks!

longlho:
should be in like 10'

PR #3081, opened 2021-07-28
Fix #3041

### LINKED ISSUE #3041: intl does not accept empty defaultMessage
## Which package?
`@formatjs/intl`

## Describe the bug
A `defaultMessage` that is explicitly an _empty string_ instead of `undefined` gets replaced by its `id` as a fallback. Messages defined in JSON to be empty work as expected. In order for the empty strings to show up correctly I had to extract and then recompile the lang file to explicitly define those IDs.

## Expected behavior
A translation's defaultMessage should be allowed to be empty. `formatjs extract` obeys this and the generated JSON has empty strings where required.

LoganDark:
Fixes #3041, take it or leave it.

### DIFF
--- a/.github/pull.yml
+++ b/.github/pull.yml
@@ -1,0 +1,5 @@
+version: "1"
+rules:
+  - base: master
+    upstream: formatjs:master
+    mergeMethod: merge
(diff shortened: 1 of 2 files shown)


LoganDark:
OK the Pull bot just screwed up the PR and I can't add the config file without that being in the PR too.

LoganDark:
Closing due to inactivity, guess my fork is here to stay. This is such a ridiculous thing to diverge over.

PR #3395, opened 2022-01-29
Removed duplicate of "t" in typescript word

joaobispo2077:
Just a grammar fix.


joaobispo2077:
Wow, I didn't know "ttypescript", I think you really mean ttypescript https://www.npmjs.com/package/ttypescript, Sorry. I suggest you add the Lib link in the word because I got confused.

PR #3853, opened 2022-09-23
feat(@formatjs/intl-numberformat): initial NumberFormat v3 implementation

ilteoood:
[Partial implementation of the NumberFormat v3](https://tc39.es/proposal-intl-numberformat-v3/out/numberformat/diff.html)

In this PR I'm not going to implement the entire NumberFormat V3 in order to avoid a huge PR.

I'll divide in the following lists what I've refactored and the new function that has been inserted.

Refactor: InitializeNumberFormat, SetNumberFormatDigitOptions, PartitionNumberPattern and prototype.resolvedOptions 
Inserted: PartitionNumberRangePattern, FormatApproximately, CollapseNumberRange, GetUnsignedRoundingMod and ApplyUnsignedRoundingMode

cc: @micheleriva


longlho:
can you add tests?

ilteoood:
Hi @longlho, tests have been inserted!

ilteoood:
Hi @longlho,
I've finished the last implementation for this PR: ApplyUnsignedRoundingMode.
I've re-executed the tests locally
image

So feel free to trigger the step.

Matteo

packages/ecma402-abstract/GetStringOrBooleanOption.ts
```diff
@@ -0,0 +1,45 @@
...
+  fallback: T[K] | boolean
+): T[K] | boolean {
+  let value: any = opts[prop]
+  if (value === undefined) {
+    return fallback
+  }
+  if (value === true) {
+    return trueValue
+  }
+  let valueBoolean = Boolean(value)
```
simoneb: nitpick: since this value is never mutated, why not use a const?

packages/ecma402-abstract/NumberFormat/InitializeNumberFormat.ts
```diff
@@ -114,34 +121,100 @@ export function InitializeNumberFormat(
...
+    options,
+    'trailingZeroDisplay',
+    'string',
+    ['auto', 'stripIfInteger'],
+    'auto'
+  )
+
+  internalSlots.trailingZeroDisplay = trailingZeroDisplay
+
+  let compactDisplay = GetOption(
```
simoneb: here also prefer const over let if the value is never mutated

simoneb (commented):
So far, so good 👌

packages/ecma402-abstract/NumberFormat/InitializeNumberFormat.ts
ilteoood: You're right, as the specification always says "let ..." sometimes I forget to convert them to const 😄

longlho:
the new v3 functionality tests seem very light. Can you add more tests for them?

longlho:
ok so Intl.NumberFormat v3 introduced `formatRange` along w/ 3 other options but in the `intl-numberformat` test I still don't see tests for them. Can you add the tests for those functionalities to that package?

ilteoood:
Hi @longlho ,
formatRange has not been implemented yet, that's the reason why you don't see any test about it.
The other options, like roundingMode, have been just declared. The functions that use them, like FormatNumericToString, need to be refactored first.

With this PR I'm not implementing the entire v3 specification, because I would like to avoid to commit too much code at once.

Please let me know if you would like to proceed in a different way.

longlho:
got it so initial goal is to not introduce regression?

packages/intl-numberformat/tests/misc.test.ts
```diff
@@ -207,8 +207,30 @@ it('chose compact pattern with rounded number', () => {
     notation: 'compact',
   })
 
-  expect(nf.format(999.995)).toEqual('1.00K')
-  expect(nf.format(999995000)).toEqual('1.00B')
+  expect(nf.format(999.995)).toEqual('1K')
```
longlho: hmm why did this change?

ilteoood: For the new logic of the SetNumberFormatDigitOptions function

ilteoood:
> Partial implementation of the NumberFormat v3

The goal here is to:
- align the refactored abstract operations to the v3 specification;
- create new abstract operations that will be needed for the v3 specification.

packages/intl-numberformat/tests/misc.test.ts
longlho: the old behavior matches with how Chrome does it
```
new Intl.NumberFormat('en', {
    // maximumIntegerDigits: 3,
    minimumFractionDigits: 2,
    maximumFractionDigits: 2,
    useGrouping: false,
    style: 'decimal',
    notation: 'compact',
  }).format(999.995)
'1.00K'
```
so this looks like a bug.

Chrome does have NFv3 implemented btw

packages/ecma402-abstract/tests/utils.ts
```diff
@@ -0,0 +1,177 @@
...
+  if (!internalSlots) {
+    internalSlots = Object.create(null) as NumberFormatInternal
+    internalSlotMap.set(x, internalSlots)
+  }
+  return {
+    ...internalSlots,
+    ...x.resolvedOptions(),
+    numberingSystem: 'latn',
+    // @ts-ignore
+    dataLocaleData: {
```
longlho, line 16: why are we hardcoding this instead of pulling from CLDR?

longlho (changes requested):

packages/ecma402-abstract/tests/utils.ts
ilteoood: I took a look at each folder of cldr, but I don't see any json file with this shape.
The jsons that I found with this shape are the one located in the `tests/locale-data` folder of the `intl-numberformat` package.

In this case, we are doing unit tests, so I would like to use this as a mock that can be easily updated.
Instead, in the number format package, there are already all the necessary json definitions from CLDR to do integration tests.

packages/intl-numberformat/tests/misc.test.ts
ilteoood: This has been fixed, my mistake.

longlho (approved):

longlho:

PR #3860, opened 2022-10-05
feat(@formatjs/intl-numberformat): implemented formatNumericRange and formatNumericRangeToParts

ilteoood:
Hi,
with this PR I implemented both formatNumericRange and formatNumericRangeToParts, with their dedicated abstract functions.

cc: @simoneb @micheleriva


packages/ecma402-abstract/NumberFormat/FormatNumericRange.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+) {
+  const parts = PartitionNumberRangePattern(numberFormat, x, y, {
+    getInternalSlots,
+  })
+
+  return parts
+    .map(part => {
+      if (part.source === 'shared') {
+        return ` ${part.value} `
```
longlho: I don't see this in the spec

packages/ecma402-abstract/NumberFormat/PartitionNumberRangePattern.ts
```diff
@@ -57,7 +57,7 @@ export function PartitionNumberRangePattern(
   const symbols =
     internalSlots.dataLocaleData.numbers.symbols[internalSlots.numberingSystem]
 
-  const rangeSeparator = symbols.timeSeparator
+  const rangeSeparator = symbols.minusSign
```
longlho: this seems wrong. It should come from https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-numbers-full/main/af/numbers.json#L142 so likely you'd have to change the data schema to handle this

longlho (changes requested):

packages/ecma402-abstract/NumberFormat/PartitionNumberRangePattern.ts
longlho: just looked at `af` and range separator is `&#8211;` while minus sign is a regular `-` so this is wrong

ilteoood: I think that I can't simply change the data schema to handle this.
Currently, symbols are fetched from a structure [like this](https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-numbers-full/main/af/numbers.json#L16); as you can see, here the `rangeSeparator` symbol is not defined.

So I've tried to hardcode that value in the [extract-numbers script](https://github.com/formatjs/formatjs/blob/main/packages/intl-numberformat/scripts/extract-numbers.ts#L64), but then lots of tests were failing due to differences between JSON test data and the new format, even in the `test262` submodule.

At the end, as this symbol is used just once in this abstract function, I've hardcoded it there.

Let me know if there is a better way to proceed.

longlho: the problem is that it's not a hardcoded symbol as I mentioned in `af` it's not `minusSign`. I think you can assumed the `rangeSeparator` comes from `range` pattern and is between `{0}` & `{1}` and write a corresponding script like `extract-numbers` and make a new file. Then you can import & use it in `core`

packages/ecma402-abstract/tests/FormatNumericRange.test.ts
```diff
@@ -0,0 +1,33 @@
...
+  it('correctly return numeric range with single negative number', () => {
+    const result = FormatNumericRange(numberFormat, -3, 10, {getInternalSlots})
+
+    expect(result).toBe('-3–10')
+  })
+
+  it('correctly return numeric range with both negative numbers', () => {
+    const result = FormatNumericRange(numberFormat, -10, -3, {getInternalSlots})
+
+    expect(result).toBe('-10–-3')
```
longlho: this seems wrong, looks like ILND is not enough because it's also dependent on the value.
Currently Chrome `formatRange(3, 10)` yields `3-10` but `formatRange(-3, -10)` yields `-3 - -10` so the spec needs to clarify this IMO.
`CollapseNumberRange` does not seem to cover this IMO
Thoughts @sffc?

longlho (changes requested):

packages/ecma402-abstract/tests/FormatNumericRange.test.ts
sffc: Good question. The UTS 35 spec is here:

http://www.unicode.org/reports/tr35/tr35-numbers.html#8-number-range-formatting

ECMA-402 does not require adherence to UTS 35, but it's a decent reference for what to do here. Regarding spacing around the range separator, it says:

> To determine whether to add spacing, the currently recommended heuristic is:
>
> 1. If the lower string ends with a character other than a digit, or if the upper string begins with a character other than a digit.
> 2. If the range pattern does not contain a character having the White_Space binary Unicode property after the `{0}` or before the `{1}` placeholders.

If both those properties are true, then add spaces. In this case, the first character of the lower string starts with a non-number character ("-"), so adding spaces is justified.

longlho: Do you think we should port this to ecma-402 or handle it in some way? Currently as the spec is written the range separator is a constant that only varies based on ILND factors and not the value itself, so inserting whitespace technically not spec-compliant.

I think a simple test262 of creating a blank Intl.NumberFormat() and formatRange on these 2 test cases I mentioned above resulting in 2 range separator would make Chrome not compliant right?

sffc: I think this is handled in `CollapseNumberRange`, which is intended to give the implementation a lot of leeway to modify the output string, but it wouldn't hurt to clarify that.

sffc: https://github.com/tc39/proposal-intl-numberformat-v3/issues/113

ilteoood: @longlho is true that if you invoke `formatRange(-3, -10)` in Chrome you obtain `-3 – -10`, but it does not seems to adhere to the specification.
Take a look [here](https://tc39.es/proposal-intl-numberformat-v3/out/numberformat/diff.html#sec-partitionnumberrangepattern): as -10 < -3, it should throw a RangeError as specified in the statement `2a`.

packages/ecma402-abstract/NumberFormat/PartitionNumberRangePattern.ts
ilteoood: The `minusSymbol` has been removed, this comment reference to an old version of the file.
Now, the `const rangeOperator` has been replaced with
```
result.push({type: 'literal', value: '–', source: 'shared'})
```

Searching inside the pattern range seems a little bit flaky, isn't it?

teobmg: ping

packages/ecma402-abstract/tests/FormatNumericRange.test.ts
longlho: image

packages/ecma402-abstract/NumberFormat/PartitionNumberRangePattern.ts
longlho: this still doesn't solve my issue, you cannot hardcode `-`

packages/ecma402-abstract/tests/FormatNumericRange.test.ts
longlho: @sffc it does seem like a spec issue with this behavior?

sffc: The range check was removed in July: https://github.com/tc39/proposal-intl-numberformat-v3/pull/100

Is the HTML version of the spec out of date? I'll fix it.

sffc: https://github.com/tc39/proposal-intl-numberformat-v3/commit/abb95a758de7bd5be069fac87879cf73d22dcc7d

packages/ecma402-abstract/NumberFormat/PartitionNumberRangePattern.ts
ilteoood: This has been addressed.

packages/ecma402-abstract/tests/FormatNumericRange.test.ts
ilteoood: I've removed all the code related to that part.

ilteoood:
Hy @longlho,
I've addressed the comments about hardcoded separator and the NumberFormat input.

The only missing one is about spacing. Should it be addressed in another PR?

longlho:
Yeah that's ok can you make a GH issue to track it?

ilteoood:
> Yeah that's ok can you make a GH issue to track it?

I opened https://github.com/formatjs/formatjs/issues/3896

longlho (approved):

longlho:

klimashkin:
Quick questions, just to understand.
This change doesn't yet introduce `formatRange` from v3, right?

Version [8.3.0](https://github.com/formatjs/formatjs/compare/@formatjs/intl-numberformat@8.2.0...@formatjs/intl-numberformat@8.3.0) says it adds `formatNumericRange`, but it's not in https://github.com/tc39/proposal-intl-numberformat-v3, so I was a bit confused

klimashkin:
Actually, seems like this adds support for `formatRange`: https://github.com/formatjs/formatjs/commit/113a169e41830ff4ddb6ae0c797292b2297bdf1d#diff-b3e50f88c036f990339b6dfd60bd482a5beb27caac9c9b84c802010974eff4da

What confused me is that `intl-numberformat/should-polyfill` is not returning `true` for me if `formatRange` is not supported, for example, in current Firefox 112 (native support [will be added in 116](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/116))

PR #3891, opened 2022-10-25
Allow `--format default` in CLI and updated documentation

CodeAndWeb:
Changes:

- Updated the documentation with a better explanation of a purely local workflow.
- Updated extraction documentation: BabelEdit supports the `default` format - including comments - this is better than using `simple`
- Adde support for `--format default` to the CLI to make it consistent with the other extraction tools. Specifying `--format default` in the current version of the CLI tool tries to read the `default` format from the current directory.

### DIFF
--- a/packages/cli-lib/src/formatters/index.ts
+++ b/packages/cli-lib/src/formatters/index.ts
@@ -35,6 +35,8 @@ export async function resolveBuiltinFormatter(
       return lokalise
     case 'crowdin':
       return crowdin
+    case 'default':
+      return defaultFormatter
   }
   try {
     // eslint-disable-next-line import/dynamic-import-chunkname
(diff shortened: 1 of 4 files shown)


longlho:
Hmm shouldn't the default just the case without the arg? Why do you need to specify default?

CodeAndWeb:
It's bit about the documentation and the usability:

What would you write in the second column for a tool that uses the default format?

Screenshot 2022-10-25 at 15 06 17

Leave it empty? In this case, people might thing that it's not supported.

If you write `default` people using `--format default` on the extract command get a confusing error message:

    Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../react-intl-demo/default' imported from .../react-intl-demo/node_modules/@formatjs/cli/bin/formatjs

This leaves the impression that "default" is not supported - and not that you just have to omit it to use it.

longlho:
you can write out something like `don't pass --format since babelEdit supports formatjs default format`

PR #3904, opened 2022-11-11
docs: fix cli npm examples

bmihelac:
from [npm](https://docs.npmjs.com/cli/v9/commands/npm-run-script#description) documentation:

> Any positional arguments are passed to the specified script. Use -- to pass --prefixed flags and options which would otherwise be parsed by npm.


pyrocat101 (approved):

pyrocat101:
Thank you for the contribution!

Issue #3930, opened 2022-12-10
Esliint rule not enforced for `$t`, an alternative to formatMessage


Oyelowo:
## Which package?
Since this is a monorepo. Please specify which package this bug report is for.
https://github.com/formatjs/formatjs/blob/main/packages/eslint-plugin-formatjs/rules/enforce-placeholders.ts

## Describe the bug
Eslint `enforce-placeholders` does not work for the short form `$t` function

## To Reproduce
```ts
function Component(){
    const { $t } = useIntl();

return (
      <div>
          {$t({ defaultMessage: "My name is {name}" })}
      </div>
  )
}
```
### Codesandbox URL

Please include a codesandbox URL if possible

### Reproducible Steps/Repo

1. Create a basic component with the example above
2. Use the short form tag. e.g: `{$t({ defaultMessage: "My name is {name}" })}`
3. Do the same for `{intl.formatMessage({ defaultMessage: "My name is {name}" })}`
4. 3 would catch the error but not 2

## Expected behavior
Eslint should catch the error when placeholder is not used when using `$t` which seems to have been added here:
https://github.com/formatjs/formatjs/commit/a0ab2d431b0bb714fd9173b7492190fa006c88be

## Screenshots
<img width="871" alt="Screenshot 2022-12-10 at 14 05 05" src="https://user-images.githubusercontent.com/31687368/206854261-fb73ec65-ff43-45d5-bfdc-55a6de4171a3.png">

## Desktop (please complete the following information):

- OS: [e.g. iOS]. MacOS
- Browser [e.g. chrome, safari] Chrome
- Version [e.g. 22].  Monterey 12.6

pyrocat101:
Fixed by #3943

PR #3943, opened 2022-12-19
fix(eslint-plugin-formatjs): adding $t, an alternative to formatMessage

cungminh2710:
This is to address #3930

Eslint should catch the error when placeholder is not used when using $t which seems to have been added here:
https://github.com/formatjs/formatjs/commit/a0ab2d431b0bb714fd9173b7492190fa006c88be

# How did I test?

`pnpm bazel test //packages/eslint-plugin-formatjs:unit-enforce-placeholders`


pyrocat101:
Thank you for your contribution!

PR #4030, opened 2023-03-19
fix(@formatjs/fast-memoize)!: use named exports

### LINKED ISSUE #3999: Can’t use `@formatjs/fast-memoize` in isomorphic code
## Which package?

`@formatjs/fast-memoize`

## Describe the bug

The `main` property of `@formatjs/fast-memoize` is a CommonJS module that specifies `module.exports.default`. There is no (real) ESM support. That means that usage in CommonJS is as follows:

```js
const memoize = require('@formatjs/fast-memoize')

memoize.default(/* … */)
```

and usage in ESM is:

```js
import memoize from '@formatjs/fast-memoize'

memoize.default(/* … */)
```

IMO this is not pretty, but it works.

The problem is that `@formatjs/fast-memoize` does specify the `module` field. This points to a **faux** ESM module. Many bundlers use this field, so correct use with a bundler is:

```js
import memoize from '@formatjs/fast-memoize'

memoize(/* … */)
```

This means it’s not possible to use `@formatjs/fast-memoize` in isomorphic code.

### Codesandbox URL

N/A

### Reproducible Steps/Repo

1. Create the following script named test.mjs:
   ```js
   import memoize from '@formatjs/fast-memoize';

   console.log(memoize);
   ``

2. Run the following script using Node.js:
   ```sh
   node test.mjs
   ```

Notice that `memoize` is a module with the property `default`. `memoize` itself is not a function.

## Expected behavior

I would expect at least one of the following:

1. `@formatjs/fast-memoize` supports proper dual packaging using package exports.
2. `@formatjs/fast-memoize` removes support for CommonJS.
3. `@formatjs/fast-memoize` removes support for faux ESM.
4. `@formatjs/fast-memoize` uses a named export instead of a default export.
5. `@formatjs/fast-memoize` gets deprecated in favour of [`fast-memoize`](https://www.npmjs.com/package/fast-memoize).

## Screenshots

N/A

- OS: Pop!_OS
- Browser: Node.js
- Version: 18

- Device: N/A
- OS: N/A
- Browser N/A
- Version N/A

## Additional context

`@formatjs/fast-memoize` was created because of a very niche issue (#1913). `fast-memoize` uses CommonJS (not UMD, as is claimed in that issue). Node.js native ESM and bundlers can consume CJS just fine. A lot of ESM packages have CJS dependencies.

pyrocat101:
Fixes https://github.com/formatjs/formatjs/issues/3999

BREAKING CHANGE: `memoize` is now a named exports rather than a default exporst to workaround shenaginans of bundler compatibility.

### DIFF
--- a/packages/fast-memoize/package.json
+++ b/packages/fast-memoize/package.json
@@ -4,8 +4,10 @@
   "description": "fork of fast-memoize and support esm",
   "main": "index.js",
   "module": "lib/index.js",
-  "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+  "exports": {
+    "import": "./lib/index.js",
+    "require": "./index.js",
+    "default": "./index.js"
   },
   "repository": {
     "type": "git",
--- a/packages/fast-memoize/index.ts
+++ b/packages/fast-memoize/index.ts
@@ -34,7 +34,7 @@ export interface MemoizeFunc<F extends Func> {
   (fn: F, options?: Options<F>): F
 }
 
-export default function memoize<F extends Func>(fn: F, options?: Options<F>) {
+export function memoize<F extends Func>(fn: F, options?: Options<F>) {
   const cache = options && options.cache ? options.cache : cacheDefault
 
   const serializer =
--- a/packages/intl-messageformat/src/core.ts
+++ b/packages/intl-messageformat/src/core.ts
@@ -9,7 +9,7 @@ import {
   MessageFormatElement,
   ParserOptions,
 } from '@formatjs/icu-messageformat-parser'
-import memoize, {Cache, strategies} from '@formatjs/fast-memoize'
+import {memoize, Cache, strategies} from '@formatjs/fast-memoize'
 import {
   FormatterCache,
   Formatters,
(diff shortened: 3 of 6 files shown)


packages/fast-memoize/package.json
```diff
@@ -4,8 +4,10 @@
   "description": "fork of fast-memoize and support esm",
   "main": "index.js",
   "module": "lib/index.js",
-  "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+  "exports": {
+    "import": "./lib/index.js",
```
remcohaszing, line 8: This breaks native ESM support. When using a native import, it resolves to this export condition. This export condition resolves to a `.js` file. Because this `package.json` doesn’t specify `"type": "module"`, this file is treated as CJS. However, it contains ESM syntax, which is a syntax error.

To fix this, this file should have a `.mjs` file extension. The accompanying type definitions should use a `.d.mts` file extension.

pyrocat101: Fixed in 2.0.1

PR #4135, opened 2023-06-19
fix(locale): migrate locale info getters to get methods

### LINKED ISSUE #4134: Rename `weekInfo` to `getWeekInfo` to be spec compliant
## Which package?

`@formatjs/intl-locale`

## Describe the bug

Currently in `intl-locale`, a getter `weekInfo` is exposed: https://github.com/formatjs/formatjs/blob/61bb35ed3a3cfc0e225077b2f5697aca85a2cee3/packages/intl-locale/index.ts#L701

But I think that according to the spec, it's supposed to be `getWeekInfo()`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo

image

## Expected behavior

The polyfill should expose a `getWeekInfo()`

Ayc0:
## Why?

Fixes https://github.com/formatjs/formatjs/issues/4134

## What does this PR do?

- Add, in addition to the getter `locale.weekInfo`, a new method `locale.getWeekInfo()` to be spec compliant , and keep `locale.weekInfo` in the polyfill
- Apply the same logic to:
  - `calendars` -> `getCalendars()`
  - `collations` -> `getCollations()`
  - `hourCycles` -> `getHourCycles()`
  - `numberingSystems` -> `getNumberingSystems()`
  - `textInfo` -> `getTextInfo()`
  - `timeZones` -> `getTimeZones()`

### DIFF
--- a/packages/intl-locale/index.ts
+++ b/packages/intl-locale/index.ts
@@ -696,9 +696,9 @@ export class Locale {
   }
 
   /**
-   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.weekInfo
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo
    */
-  public get weekInfo() {
+  public getWeekInfo() {
     try {
       const info = Object.create(Object.prototype)
       const wi = weekInfoOfLocale(this)
@@ -729,6 +729,14 @@ export class Locale {
     } catch (e) {
       throw new TypeError('Error retrieving weekInfo')
     }
+  }
+
+  /**
+   * @deprecated – use local.getWeekInfo() instead
+   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.weekInfo
+   */
+  public get weekInfo() {
+    return this.getWeekInfo()
   }
 
   static relevantExtensionKeys = RELEVANT_EXTENSION_KEYS
--- a/packages/intl-locale/index.ts
+++ b/packages/intl-locale/index.ts
@@ -640,44 +640,90 @@ export class Locale {
   }
 
   /**
-   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.calendars
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getCalendars
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getCalendars
    */
-  public get calendars() {
+  public getCalendars() {
     return calendarsOfLocale(this)
   }
 
   /**
-   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.collations
+   * @deprecated – use local.getCalendars() instead
+   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.calendars
    */
-  public get collations() {
+  public get calendars() {
+    return this.getCalendars()
+  }
+
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getCollations
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getCollations
+   */
+  public getCollations() {
     return collationsOfLocale(this)
   }
 
   /**
-   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.hourCycles
+   * @deprecated – use local.getCollations() instead
+   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.collations
    */
-  public get hourCycles() {
+  public get collations() {
+    return this.getCollations()
+  }
+
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getHourCycles
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getHourCycles
+   */
+  public getHourCycles() {
     return hourCyclesOfLocale(this)
   }
 
   /**
-   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.numberingSystems
+   * @deprecated – use local.getHourCycles() instead
+   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.hourCycles
    */
-  public get numberingSystems() {
+  public get hourCycles() {
+    return this.getHourCycles()
+  }
+
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getNumberingSystems
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getNumberingSystems
+   */
+  public getNumberingSystems() {
     return numberingSystemsOfLocale(this)
   }
 
   /**
-   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.timeZones
+   * @deprecated – use local.getNumberingSystems() instead
+   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.numberingSystems
    */
-  public get timeZones() {
+  public get numberingSystems() {
+    return this.getNumberingSystems()
+  }
+
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTimeZones
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getTimeZones
+   */
+  public getTimeZones() {
     return timeZonesOfLocale(this)
   }
 
   /**
-   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.textInfo
+   * @deprecated – use local.getTimeZones() instead
+   * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.timeZones
    */
-  public get textInfo() {
+  public get timeZones() {
+    return this.getTimeZones()
+  }
+
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTextInfo
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getTextInfo
+   */
+  public getTextInfo() {
     try {
       const info = Object.create(Object.prototype)
       const dir = characterDirectionOfLocale(this)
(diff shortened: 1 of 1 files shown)


Youpinadi (approved):

Ayc0:
@pyrocat101 what do you think about this?

I also noticed the same thing with `hourCycles` and `getHourCycles` (https://github.com/formatjs/formatjs/blob/61bb35ed3a3cfc0e225077b2f5697aca85a2cee3/packages/intl-locale/index.ts#L659)

I can include it in this PR too

packages/intl-locale/index.ts
```diff
@@ -640,44 +640,90 @@ export class Locale {
...
   /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getCalendars
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getCalendars
+   */
+  public getCalendars() {
+    return calendarsOfLocale(this)
+  }
+
+  /**
+   * @deprecated – use local.getCalendars() instead
```
longlho: this is not deprecated. Same as the ones below

packages/intl-locale/index.ts
```diff
@@ -640,44 +640,90 @@ export class Locale {
...
    * https://tc39.es/proposal-intl-locale/#sec-Intl.Locale.prototype.calendars
    */
   public get calendars() {
-    return calendarsOfLocale(this)
+    return this.getCalendars()
```
longlho: you should revert this change. This applies to the below implementations as well

longlho (changes requested):

packages/intl-locale/index.ts
Ayc0: According the MDN:

> Note: In some versions of some browsers, this method was implemented as an accessor property called `calendars`. However, because it returns a new array on each access, it is now implemented as a method to prevent the situation of locale.calendars === locale.calendars returning false. Check the browser compatibility table for details.

This seems to imply that using the getter `.calendars` is deprecated

packages/intl-locale/index.ts
Ayc0: If both `.calendars`  and `.getCalendars()` are exposed, shouldn't one use the other?

packages/intl-locale/index.ts
longlho: The spec doesn't say it's deprecated

packages/intl-locale/index.ts
longlho: No that's not what the spec says

packages/intl-locale/index.ts
Ayc0: They fully removed every mention of the old getters (on https://tc39.es/proposal-intl-locale-info, the getter `.calendars` is not referenced anymore, just `.getCalendars` is)

I only kept the getters to avoid introducing a major change and to be able to tell users of the lib that they should switch to the new attributes

longlho: My previous point still stands. Spec doesn't say deprecated anywhere.

Ayc0: What are you recommending me to do on this PR?
1. Keep both `.getCalandars` and `.calendars` as it is at the moment
2. Only keep the new attributes and remove the getter ones (to follow the new spec)
3. Remove the deprecation notice but keep the rest unchanged (2 attributes with the getters calling the other ones)
4. Keep both without calling `this.getCalendars()` in `get calendars` (and duplicating the logic)
5. Only keep the getters and removing the new methods (aka closing this PR)
6. Another path

longlho: 4 sounds like the most spec compliant option so I'd recommend that

longlho: None of the other options would pass test262

Ayc0: Updated

packages/intl-locale/index.ts
```diff
@@ -639,41 +639,103 @@ export class Locale {
...
   public get timeZones() {
     return timeZonesOfLocale(this)
   }
 
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTextInfo
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getTextInfo
+   */
+  public getTextInfo() {
+    try {
```
longlho: this try catch is not part of the spec

packages/intl-locale/index.ts
```diff
@@ -695,6 +757,43 @@ export class Locale {
     }
   }
 
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getWeekInfo
+   */
+  public getWeekInfo() {
+    try {
```
longlho: not in spec

packages/intl-locale/index.ts
```diff
@@ -695,6 +757,43 @@ export class Locale {
...
+      const info = Object.create(Object.prototype)
```
longlho: missing `initializedLocale` check per spec

packages/intl-locale/index.ts
```diff
@@ -639,41 +639,103 @@ export class Locale {
...
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTextInfo
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getTextInfo
+   */
+  public getTextInfo() {
+    try {
+      const info = Object.create(Object.prototype)
+      const dir = characterDirectionOfLocale(this)
+
+      Object.defineProperty(info, 'direction', {
```
longlho: please use `defineProperty` from `@formatjs/ecma402-abstract`. Same as below

packages/intl-locale/index.ts
```diff
@@ -639,41 +639,103 @@ export class Locale {
...
   public get collations() {
     return collationsOfLocale(this)
   }
 
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getHourCycles
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getHourCycles
+   */
+  public getHourCycles() {
+    return hourCyclesOfLocale(this)
```
longlho, line 683: missing `initializedLocale` check per spec

packages/intl-locale/index.ts
```diff
@@ -639,41 +639,103 @@ export class Locale {
...
   public get numberingSystems() {
     return numberingSystemsOfLocale(this)
   }
 
+  /**
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getTimeZones
+   * https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getTimeZones
+   */
+  public getTimeZones() {
+    return timeZonesOfLocale(this)
```
longlho, line 713: this is missing canonical locale check per spec

longlho (changes requested):

packages/intl-locale/index.ts
Ayc0: I don't fully understand what you mean by "not in the spec":  in https://tc39.es/proposal-intl-locale-info/#sec-Intl.Locale.prototype.getWeekInfo, it's mentioned `Perform ! CreateDataPropertyOrThrow` which itself says "If success is false, throw a TypeError exception"

image

I didn't write this original code (as I copy pasted it from `public get weekInfo()`) but if I understand this code properly this `try {} catch {}` is for changing all possible errors into `TypeError` as mentioned in the spec.

Can you help me on how this code should be adapted?

packages/intl-locale/index.ts
Ayc0: I'm not sure I understand what this means:

In the spec here, I don't see any `initializedLocale` check:
image

Is it the `RequireInternalSlot`?

Or is it the step `Assert: locale matches the unicode_locale_id production.` within `WeekInfoOfLocale(loc)`?

image

packages/intl-locale/index.ts
Ayc0: Same as the other comment, I'm not sure I understand what you mean by that (as I'm not that familiar with the exact spec). Can you explain to me what this is?

packages/intl-locale/index.ts
Ayc0: Thanks for the feedback 🙇 

I updated this `Object.defineProperty` and the ones in `getWeekInfo` too

packages/intl-locale/index.ts
longlho: for this one you can just remove the `try/catch`. `defineProperty` & `CreateDataPropertyOrThrow` both call `DefinePropertyOrThrow` so that should already be covered

packages/intl-locale/index.ts
longlho: yeah per spec you basically need to check `initializedLocale` in `internalSlots` to be set/defined before proceeding

packages/intl-locale/index.ts
longlho: ok this one looks good looks like it's already covered in the impl of `timeZonesOfLocale`
```
const region = parseUnicodeLanguageId(locale).region

  if (!region) {
    return undefined
  }
  ```

packages/intl-locale/index.ts
Ayc0: Updated 👍

packages/intl-locale/index.ts
Ayc0: Updated 👍

packages/intl-locale/index.ts
Ayc0: Isn't the `initializedLocale` ensured by the constructor of the Locale?

image

Ayc0: Also if you check `baseName`, there is the same `RequireInternalSlot` in the spec:
image

But the implem only calls `getInternalSlots(this)`:

image

Which is also what `weekInfoOfLocale` does internally:
image
image

So I'm not sure what I need to add to this method

packages/intl-locale/index.ts
Ayc0: Same as the comment above: `hourCyclesOfLocale` is using internally `getInternalSlots` so I'm not sure where I should add this `initializedLocale` check

image

packages/intl-locale/index.ts
longlho: I think you basically just need to make sure `initializedLocale` is not `undefined` in `internalSlots`, otherwise throw per `RequireInternalSlot` spec

Ayc0: Added 👍  (I didn't added it to `weekInfoOfLocale` to avoid impacting the getter `weekInfo`)

packages/intl-locale/index.ts
Ayc0: Added 👍

longlho (approved):

longlho:

PR #4206, opened 2023-10-12
Update distribute-libraries.md

gokatz:
fix possible typo


gokatz:
looks like there is an existing PR to fix this

PR #4345, opened 2024-02-02
fix(@formatjs/cli-lib): onMsgExtracted / onMetaExtracted not getting called

### LINKED ISSUE #4343: cli-lib: onMsgExtracted never gets called
## Which package?

cli-lib 

## Describe the bug

Setting the option `onMsgExtracted` has no effect when using `extract` from `@formatjs/cli-lib`, the callback never gets called because that option gets overridden here: https://github.com/formatjs/formatjs/blob/34634cec8d84c7cee4ba4e80e1384c758c67ede1/packages/cli-lib/src/extract.ts#L106

### Codesandbox URL

n/a

### Reproducible Steps/Repo

1. Run a script like:

```
import { extract } from "@formatjs/cli-lib";
import { DEFAULT_ID_INTERPOLATION_PATTERN } from "babel-plugin-formatjs";

await extract(<files>, {
  extractSourceLocation: true,
  // needed for some reason
  idInterpolationPattern: DEFAULT_ID_INTERPOLATION_PATTERN,
  throws: true,
      
  // doesn't get called 🤥
  onMsgExtracted: (...args) => {
    console.log("onMsgExtracted");
  },
});
```

2. Observe that `onMsgExtracted` doesn't get printed.

## Expected behavior

`onMsgExtracted` should be printed on the console.

n/a

n/a

## Additional context

Nothing I can think of.

adbl:
Call `onMsgExtracted` and `onMetaExtracted` options if passed as options to `extract` in cli-lib. Fixes #4343

### DIFF
--- a/packages/cli-lib/src/extract.ts
+++ b/packages/cli-lib/src/extract.ts
@@ -97,13 +97,15 @@ async function processFile(
   let messages: ExtractedMessageDescriptor[] = []
   let meta: Record<string, string> | undefined
 
+  const onMsgExtracted = opts.onMsgExtracted
+
   opts = {
     ...opts,
     additionalComponentNames: [
       '$formatMessage',
       ...(opts.additionalComponentNames || []),
     ],
-    onMsgExtracted(_, msgs) {
+    onMsgExtracted(filePath, msgs) {
       if (opts.extractSourceLocation) {
         msgs = msgs.map(msg => ({
           ...msg,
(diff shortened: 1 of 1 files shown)


adbl:
@longlho thanks for quick review! May I request a new release of the package as we depend on this in an ongoing project.

adbl:
@longlho would it be possible to initiate a patch release for `cli-lib`, it would simplify and unblock our ongoing project here.

PR #4376, opened 2024-03-18
feat(cli-lib): adding support for gts, gjs and hbs files.

kiwiupover:
## Use case

Engineering using [glimmerjs](https://glimmerjs.com/) files `.gts` and `.gjs` and handlebar files `.hbs` want to use formatjs for internationalization  

## Solution

Follow the Vue example and use [glimmer](https://github.com/glimmerjs/) tooling and [ember-template-recast](https://github.com/ember-template-lint/ember-template-recast) to manage code transformation.  

Thanks to @mansona and @candunaj for the initial work.


package.json
```diff
@@ -86,6 +90,8 @@
     "clsx": "2",
     "commander": "8",
     "core-js": "^3.6.5",
+    "ember-template-recast": "^6.1.4",
+    "ember-template-tag": "^2.3.16",
```
NullVoxPopuli:
```suggestion
    "content-tag": "^2.0.0",
```

`content-tag` is the official implementation of `<template>` parsing (and has types!)

See here: https://github.com/embroider-build/content-tag/

packages/cli-lib/package.json
```diff
@@ -33,12 +33,19 @@
...
+    "@glimmer/syntax": "^0.89.0",
+    "@glimmer/validator": "^0.89.0",
     "@types/estree": "^1.0.0",
     "@types/fs-extra": "^9.0.1",
     "@types/json-stable-stringify": "^1.0.32",
     "@types/node": "14 || 16 || 17",
     "chalk": "^4.0.0",
     "commander": "8",
+    "ember-template-recast": "^6.1.4",
+    "ember-template-tag": "^2.3.16",
```
NullVoxPopuli: same here

packages/cli-lib/src/gts_extractor.ts
```diff
@@ -0,0 +1,19 @@
...
+  const scriptParseFn = parseScript(options, fileName)
+  const {output: transformedSource} = transform({
+    input: source,
+    relativePath: '',
+  })
+
+  scriptParseFn(transformedSource)
+
+  // extract template from transformed source to then run through hbs processor
+  const [templateSource] = parseTemplates(source, '')
```
NullVoxPopuli: gts can have multiple templates, so does this concatenate all of the different templates?

For example, looking at this type from `content-tag`: https://github.com/embroider-build/content-tag/?tab=readme-ov-file

We have:
```js
import { Preprocessor } from 'content-tag';
let p = new Preprocessor();

// ...
let output = p.parse(source);

for (let parseResult of output) {
  parseHbsFile(parseResult.contents, filename, options);
}
```

packages/cli-lib/tests/unit/fixtures/comp.gts
```diff
@@ -0,0 +1,69 @@
+import formatMessage from 'ember-intl/helpers/format-message';
+import or from 'ember-truth-helpers/helpers/or';
+import Component from '@glimmer/component';
+import {service} from '@ember/service';
+
```
NullVoxPopuli, line 5: I think if a second component were added to this file, the implementation (as is) would break -- unless I misunderstanding something about how `ember-template-tag` works opposed to `content-tag`

packages/cli-lib/src/hbs_extractor.ts
```diff
@@ -0,0 +1,31 @@
+import {transform} from 'ember-template-recast'
+
+function extractText(node: any, fileName: any, options: any) {
```
longlho: do those params have real types?

longlho:
overall this needs more typings

longlho (changes requested):

packages/cli-lib/src/hbs_extractor.ts
NullVoxPopuli: they do, 
```ts
import type { AST } from '@glimmer/syntax';

function extractText(node: AST.MustacheStatement | AST.SubExpression, ...) {}
```

package.json
NullVoxPopuli: Resolved!

packages/cli-lib/src/hbs_extractor.ts
NullVoxPopuli: types have been resolved

NullVoxPopuli:
atm, I'm having issues with ESM compatibility in this CJS project hmmm

NullVoxPopuli:
it looks like esbuild is what's used to compile, and it _does_ support the situation we have

[REPL here](https://hyrious.me/esbuild-repl/?version=0.19.10&b=e%00entry.js%00async+function+start%28%29+%7B%0A++await+import%28%27.%2Fexample.mjs%27%29%3B%0A%7D%0A%0Astart%28%29%3B&b=%00example.mts%00console.log%28%27hi%27%29&o=--format%3Dcjs+--target%3Dnode18+--platform%3Dnode)

so, something else is goofy in formatjs :thinking: 
The error:
```
✘ [ERROR] Could not resolve "./hbs_extractor.mts"

    ../../../../../../../../execroot/formatjs/bazel-out/k8-fastbuild/bin/packages/cli-lib/lib_esnext/src/extract.js:76:43:
      76 │         const { parseFile } = await import('./hbs_extractor.mts');
         ╵                                            ~~~~~~~~~~~~~~~~~~~~~

✘ [ERROR] Could not resolve "./gts_extractor.mts"

    ../../../../../../../../execroot/formatjs/bazel-out/k8-fastbuild/bin/packages/cli-lib/lib_esnext/src/extract.js:84:43:
      84 │         const { parseFile } = await import('./gts_extractor.mts');
         ╵                                            ~~~~~~~~~~~~~~~~~~~~~

```

it seems a combination of typescript, esbuild, buildtime and runtime are all in disagreement with eachother with what valid imports are

NullVoxPopuli:
this might be blocked on moving away from jest? 

I'm running just the test command, and seeing: 
```
❯   npx bazel run //packages/cli-lib:unit_test


  ● hbs_extractor

    Cannot find module '../../src/hbs_extractor.mjs' from 'tests/unit/hbs_extractor.test.ts'

      4 |
      5 | test('hbs_extractor', async function() {
    > 6 |   const { parseFile } = await import('../../src/hbs_extractor.mjs');
        |                               ^
      7 |   let messages: MessageDescriptor[] = []
      8 |   const fixturePath = join(__dirname, './fixtures/comp.hbs')
      9 |   parseFile(await readFile(fixturePath, 'utf8'), fixturePath, {

      at Resolver._throwModNotFoundError (../../node_modules/.aspect_rules_js/jest-resolve@29.5.0/node_modules/jest-resolve/build/resolver.js:427:11)
      at tests/unit/hbs_extractor.test.ts:6:31

```

longlho:
we don't support `.mts` or `.mjs`. The interop is a pain to deal with

NullVoxPopuli:
It's pretty straight forward without jest, tbh. But i'll have another go at by pushing the module boundary into the functions, and make them all async (we previously had everything passing except for type checking, which is how things got to how they are now).

I'll see if i can get this repo's ts config, jest, and esbuild all in agreement that way! 🎉

Also, in a separate pr, i can try moving this repo to vitest for ya, if that's of interest?

longlho:
Thanks for your contribution. Yeah I don't have a strong opinion on `vitest` vs `jest`, so if u wanna make a PR that'd be great as well

longlho:
hmm getting vitest to work can be pretty hard, any reason you can't use just `.ts` instead of `.mts`?

NullVoxPopuli:
> can't use just .ts instead of .mts?

I cannot due to type checking failing. We can't require or await import (for some reason) from cjs(ts), even though it's allowed generally, it's just type checking that is failing, iirc.

I discovered that this is exclusively a tsc problem tho. If i use vite for the build, as shown, everything is fine! 🎉 
Same with vitest (in cli-lib) -- it's already working flawlessly!

The last step i have to figure out is how to have cli-lib's different tools integrated into bazel

longlho:
what's failing typechecking? This PR now has tooling changes so I can't really tell from its logs

longlho:
from all the commits you pushed which one doesn't have the tooling changes?

NullVoxPopuli:
> what's failing typechecking? This PR now has tooling changes so I can't really tell from its logs

If you revert the most recent commit, you'll see

NullVoxPopuli:
I don't know what's going on with the build anymore, getting errors like this:
image

These are SWC, but there is also ESBuild and TSC in use. Why so many compilers? :sweat_smile:

longlho:
@NullVoxPopuli I've manually merged this in after a bunch of patches. I basically remove `type: module` from glimmer packages & `content-tag` so things would work. `mjs` overall is a mess and the ecosystem's not ready for it. If you wanna add integration tests to make sure parsing `hbs`/`gts` works that'd be great too.

longlho:
I've added the integration tests in https://github.com/formatjs/formatjs/commit/fd1e34725fbb480af684b774cc8f743e39c86bf5#diff-0ead01fee706ec365313e79e65339cfa742bf05b2f05587d1e30815195fed00b
looks like hbs works but not gts/gjs

longlho:
looks like packaging issue `Error: ENOENT: no such file or directory, open '/private/var/tmp/_bazel_long.ho/a2216e06a77a9425f6dcd5fd195f64d5/execroot/formatjs/bazel-out/darwin_arm64-fastbuild/bin/packages/cli/bin/content_tag_bg.wasm'`

longlho:
ok latest version should fix this. Integration tests also passed. Releasing that now.

NullVoxPopuli:
Thanks for solving everything! Sorry I couldn't figure it out (i had full esm and cjs compat easily if it weren't for jest (vitest 'just works')). the ESM/CJS compat I've don is much simpler, but is also only using one cohesive build tool :sweat_smile: 
maybe I just didn't spend enough time with it, or am not smart enough. haha.

You are my hero, today @longlho <3

longlho:
No worries. I know you contribute upstream so if there're things to make it easier to not have to patch on my side that'd be great :)

kiwiupover:
@longlho thanks heaps for your work getting `.gts`, `.gjs` and `.hbs` into formatjs. you have made the ember community happy, mate. Cheers.

longlho:
What linter do you guys use? Extractor is only 1 piece but we also have linter integration for react/vue as well.

NullVoxPopuli:
do you mean translations linter? (I'm a newbie to the tooling side of this space (translations), fwiw :sweat_smile: )
for js/ts/etc linting, we use eslint and prettier

longlho:
What's the eslint plugin for those extensions & do they use the same parser?
For capabilities you can look at existing formatjs linter at formatjs.io

NullVoxPopuli:
ah! yes!

For ESLint we use:
 - https://github.com/ember-cli/eslint-plugin-ember

which uses:
 - https://github.com/NullVoxPopuli/ember-eslint-parser

which takes our gjs/gts parser, content-tag:
- https://github.com/embroider-build/content-tag/

And stitches together ESTree AST and uses our `@glimmer/syntax` parser 
- https://github.com/glimmerjs/glimmer-vm/tree/main/packages/%40glimmer/syntax

and prefixes _those_ AST nodes with `Glimmer`
So we have `GlimmerElementNode`, `GlimmerAttrNode`, etc

We still need to do the same for jscodeshift -- using a custom parser that is actually two parsers -- but that is pending on some cleanup work I need to get back to here: https://github.com/embroider-build/content-tag/pull/75
Which enables eslint, prettier, jscodeshift and any other plugin/ecosystem to have better control over offsets and such with the gjs/gts <-> pure JS transformation.

For Prettier, all we do use this plugin:
- https://github.com/gitKrystan/prettier-plugin-ember-template-tag/

which also uses content-tag for parsing, but because prettier already supported formatting glimmer, we use invoke prettier on each `<template>` block, handle the indentation offset per line, and stitch the file back together.

All of this is pretty similar to how Svelte and Vue work, afaik -- but _also_ afaik, we're the only ones more eagerly integrating ESM into our CJS tooling (and fairly seamlessly!)

----------

- `content-tag` supports both ESM and CJS (it is cross-compiled)
- `@glimmer/syntax` supports both ESM and CJS (it is cross-compiled)
- ESLint plugins are CJS
- Prettier plugins are ESM
- Our [template-lint](https://github.com/ember-template-lint/ember-template-lint) tool and its plugins are ESM

longlho:
Do you guys use ember-intl or formatjs directly? I was trying to get ember-intl merged into this monorepo but that doesn't pan out. That package is pretty old as well so thinking of just forking it.

NullVoxPopuli:
the community makes a lot of use of ember-intl, and my employer does as well, but we were thinking about convincing (or really making a solid case for) ember-intl to be a build-time plugin rather than more runtime focused as it is now -- at work, we don't do much special in our product code to author translations -- we extract all the strings and such at build time, and only do manual / translating if we have tokens/numbers/data/etc involved that end up splicing up the same translation string. This makes for a very good default experience where the source code is just authored in english, and we don't maintain translation files or anything.

To do this, we use formatjs directly in our build time code.

longlho:
Can you describe that pipeline in a bit more details?

PR #4458, opened 2024-06-04
Documented how to use ember-intl

ijlee2:
Hello, everyone. In https://github.com/ember-intl/ember-intl/issues/1801, @longlho had reached out to see if it might be possible to move the `ember-intl` package to this repo. I had declined the offer and proposed, as an alternative, a larger mention of `ember-intl` in the `formatjs` website, so that people from the wider JS community may learn how to set up and use `ember-intl`.

In case you (and possibly other maintainers) are still open to this idea of updating the `formatjs` documentation, I created this pull request so that we can begin a conversation. Thanks in advance.

Preview:

### DIFF
--- a/website/docusaurus.config.ts
+++ b/website/docusaurus.config.ts
@@ -33,7 +33,7 @@ export default {
           label: 'API References',
           position: 'left',
           activeBaseRegex:
-            'docs/(vue-intl|intl|react-intl|intl-messageformat|icu-messageformat-parser)',
+            'docs/(vue-intl|intl|ember-intl|react-intl|intl-messageformat|icu-messageformat-parser)',
         },
         {
           to: 'docs/polyfills',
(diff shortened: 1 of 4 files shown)


website/docusaurus.config.ts
```diff
@@ -33,7 +33,7 @@ export default {
           label: 'API References',
           position: 'left',
           activeBaseRegex:
-            'docs/(vue-intl|intl|react-intl|intl-messageformat|icu-messageformat-parser)',
+            'docs/(vue-intl|intl|ember-intl|react-intl|intl-messageformat|icu-messageformat-parser)',
```
ijlee2, line 36: I wasn't sure what `activeBaseRegex` does. I updated the line because it had mentioned `vue-intl` and `react-intl`.

ijlee2:
Hi, @longlho. Would you be interested in this PR getting merged, so that the documentation provides more information about using `formatjs` with Ember?

PR #4479, opened 2024-07-25
docs: Use `locale` in import path for "Dynamic Import" example code

romellem:
I think this is just a typo.

The `unsupportedLocale` var is used above, but it is a boolean, not the `locale` string that is passed in these examples.

For example, `intl-displaynames.md` has the correct example code:

https://github.com/formatjs/formatjs/blob/f71ce48aaf8ada6a77d393d3f2293fc14f2c38c2/website/docs/polyfills/intl-displaynames.md#L78


romellem:
🤦‍♂️ nevermind, I was using an older version of formatjs (v5) which didn't support this. That code will work on the latest version, carry on!

PR #4700, opened 2024-11-14
fix(@formatjs/cli-lib): instantiate FormatFn with unknown in Formatter definition

pablobirukov:
Whenever we use `FormatFn` without explicit type argument, it's being instantiated with default type argument as follows
```
(msgs: Record<string, MessageDescriptor>) => Record<string, MessageDescriptor>
``` 
which is not correct in many places and lead to pretty much any custom formatter being non-assignable to `Formatter` interface:

image

We probably should instantiate it with `any` or `unknown` because we actually do not care about the return type and leave it up to a user.

### DIFF
--- a/packages/cli-lib/src/extract.ts
+++ b/packages/cli-lib/src/extract.ts
@@ -69,7 +69,7 @@ export type ExtractOpts = Opts & {
   /**
    * Path to a formatter file that controls the shape of JSON file from `outFile`.
    */
-  format?: string | Formatter
+  format?: string | Formatter<unknown>
   /**
    * Whether to hoist selectors & flatten sentences
    */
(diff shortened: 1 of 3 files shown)


packages/cli-lib/src/extract.ts
```diff
@@ -217,7 +217,9 @@ export async function extract(
     }
   }
 
-  const formatter: Formatter = await resolveBuiltinFormatter(opts.format)
+  const formatter: Formatter<unknown> = await resolveBuiltinFormatter(
```
pablobirukov, line 220: Previously, `formatter` was instantiated
```
{
  serialize?: (msgs: Record<string, MessageDescriptor>) => string
  format: (msgs: Record<string, MessageDescriptor>) => Record<string, MessageDescriptor>
  compile: (msgs: Record<string, MessageDescriptor>) => Record<string, string>
  compareMessages?: Comparator
}
```
for every formatter, while in runtime it was different. E.g., for `smartling`:
```
{
  serialize?: (msgs: SmartlingJson) => string
  format: (msgs: Record<string, MessageDescriptor>) => SmartlingJson
  compile: (msgs: SmartlingJson) => Record<string, string>
  compareMessages?: Comparator
}
```
etc

So, having less type information is more correct in that case

longlho (approved):

longlho (approved):

Issue #4501, opened 2024-08-30
Sharable eslint rule config


cungminh2710:
[ESLint shareable configurations](https://eslint.org/docs/latest/developer-guide/shareable-configs) exist to provide a comprehensive list of rules settings that you can start with.

It would be nice to predefine a sharable configuration for developers to start using the linter quickly with recommended practices.

I created this ticket so that I can remember and work on it later

## Which package?

eslint-plugin-formatjs

longlho:
yeah I'd be down for that

cungminh2710:
will work on this soon

michaelfaith:
@cungminh2710 still thinking of doing this?

cungminh2710:
Hi @michaelfaith , I'd been busy with work and on parental leave now. Didn't have a chance to have a go. 

I was thinking of something like this 

```
rules: {
    "formatjs/no-invalid-icu": "error",
    "formatjs/enforce-description": "error",
    "formatjs/enforce-default-message": ["error", "literal"],
    "formatjs/enforce-placeholders": "error",
    "formatjs/no-id": "error",
    "formatjs/no-emoji": "error",
    "formatjs/no-useless-message": "error",
    "formatjs/no-multiple-plurals": "error",
    "formatjs/no-offset": "error",
    "formatjs/blocklist-elements": [2, ["selectordinal"]],
    "formatjs/no-complex-selectors": "error",
    "formatjs/no-missing-icu-plural-one-placeholders": "error"
  },
```

michaelfaith:
@longlho This isn't compatible with the new flat config format, it would only work with the deprecated legacy/rc configs.  For instance, the `plugin` entry can't be set as a string anymore https://github.com/formatjs/formatjs/commit/a4fc4379700283448350b6982338f169f15d9bba#diff-f106795198066833ac5475a8edd7489760647e7587a4d76812c9120b8595c4c5R70

Happy to help correct it, if you need.

HarrisonLeach1:
@michaelfaith, please do. Our team needs it too 🙏

michaelfaith:
I'll try and have a PR up sometime tomorrow.

michaelfaith:
Issue: https://github.com/formatjs/formatjs/issues/4705
PR: https://github.com/formatjs/formatjs/pull/4706

PR #4706, opened 2024-11-17
fix(eslint-plugin-formatjs): make shareable configs compatible with flat config

### LINKED ISSUE #4705: [eslint-plugin] New Shareable Configs Are Not Compatible
## Which package?

eslint-plugin-formatjs

## Describe the bug

The new shareable configs (introduced by https://github.com/formatjs/formatjs/issues/4501) are in the deprecated legacy/rc format, not the new flat config format.  Most notably, the `plugins` key is an object now and not a string array.  They also need the `meta` properties.  And since the plugin as of v5 only supports eslint v9, it doesn't make sense to only support the deprecated config format. 

## To Reproduce

Use either of the new configs with eslint v9, and you'll get an error about the config format being incompatible.

michaelfaith:
The new shareable configs introduced by #4501 were in the deprecated legacy/rc format, not the new flat config format. Most notably, the plugins key is an object now and not a string array. They also needed the meta properties. And since the plugin as of v5 only supports eslint v9, it doesn't make sense to only support the deprecated config format.

This change addresses those issues, making the configs work (only) with the new format.

Fixes #4705


longlho:
thanks for putting this together. Is there a way to write test for this?

michaelfaith:
> thanks for putting this together. Is there a way to write test for this?

What i generally do for testing the configs, is more in the form of e2e tests. Have an "examples" folder with a set of simple projects that use the plugin and run lint as part of ci testing. Another benefit of that is it acts as a sort of documentation you can point people to for usage examples.  What do you think?

michaelfaith:
Added `resolveJsonModule` to `tsconfig.node.json` so that the `package.json` import would work.

longlho:
hmm ill take it from here.

packages/eslint-plugin-formatjs/index.ts
```diff
@@ -64,119 +64,131 @@ import {
...
+        {
+          props: {
+            include: [['*', '{label,placeholder,title}']],
           },
-        ],
-        'formatjs/blocklist-elements': ['error', {elements: ['selectordinal']}],
-      },
+        },
+      ],
+      'formatjs/blocklist-elements': ['error', {elements: ['selectordinal']}],
```
michaelfaith: Oo, the integration tests served their purpose.  Looks like there was an issue with the original config for `blocklist-elements`
```suggestion
      'formatjs/blocklist-elements': ['error', ['selectordinal']],
```
Per
image

packages/eslint-plugin-formatjs/index.ts
```diff
@@ -64,119 +64,131 @@ import {
...
```
michaelfaith:
```suggestion
      'formatjs/blocklist-elements': ['error', ['selectordinal']],
```

packages/eslint-plugin-formatjs/index.ts
michaelfaith: Also that screen shot made me realize that the documentation on the site should probably be updated for the new config format

longlho:
hmm that test passes which is a little odd cause that file definitely doesn't conform to `strict`. Maybe the exit code for `pnpm` is wack

longlho:
oh it's because it's a typescript file hmm. Now it fails which is what I expected

longlho (approved):

longlho:

PR #4954, opened 2025-04-03
Update type of the <FormattedNumber> `value` prop

probertson:
It was hard-coded as `number`. The `formatNumber()` function has been updated to accept [all values that `Intl.NumberFormat.format()` accepts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/format#number), so this makes the component match the underlying function.

The component already works correctly if you pass (for example) a numeric string as the `value` prop. This updates the type checking to match the reality of what the component supports.

### DIFF
--- a/packages/react-intl/index.ts
+++ b/packages/react-intl/index.ts
@@ -27,6 +27,7 @@ import FormattedPlural from './src/components/plural'
 import IntlProvider from './src/components/provider'
 import FormattedRelativeTime from './src/components/relative'
 import useIntl from './src/components/useIntl'
+import {IntlShape} from './src/types'
 export {
   createIntlCache,
   CustomFormatConfig,
@@ -94,7 +95,7 @@ export const FormattedTime: React.FC<
 export const FormattedNumber: React.FC<
   Omit<NumberFormatOptions, 'localeMatcher'> &
     CustomFormatConfig<'number'> & {
-      value: number
+      value: Parameters<IntlShape['formatNumber']>[0]
       children?(formattedNumber: string): React.ReactElement | null
     }
 > = createFormattedComponent('formatNumber')


packages/react-intl/index.ts
```diff
@@ -94,7 +95,7 @@ export const FormattedTime: React.FC<
 export const FormattedNumber: React.FC<
   Omit<NumberFormatOptions, 'localeMatcher'> &
     CustomFormatConfig<'number'> & {
-      value: number
+      value: Parameters<IntlShape['formatNumber']>[0]
```
probertson, line 98: I made this use generics in order to keep it in sync with @formatjs/intl::formatNumber. However, I noticed a comment on line 80 saying "Explicit here to prevent api-extractor from outputing `import('./src/types')`..."

I wasn't sure if that comment also applies to this line. If it does, I propose changing it to make it match the type of the `value` argument of [@formatjs/intl::formatNumber](https://github.com/formatjs/formatjs/blob/d367c853e189aae48b6b1dafd5c5e35ab62cadcd/packages/intl/src/number.ts#L71) -- namely:
```
value: Parameters<Intl.NumberFormat['format']>[0]
```
or alternatively, from [the `intl-numberformat` polyfill](https://github.com/formatjs/formatjs/blob/d367c853e189aae48b6b1dafd5c5e35ab62cadcd/packages/intl-numberformat/src/types.ts#L16):
```
value: number | bigint | Decimal | string
```

probertson:
I'm also unsure how to get the test pipeline to work. I don't _think_ it's related to my changes -- it seems like a permissions issue (?) -- but if I'm wrong I'd be happy to try to fix it.

PR #4979, opened 2025-05-07
fix(@formatjs/cli-lib): main is to point to lib_esnext/

pablobirukov:
Attempting to install 7.4.1.

Build results under is `lib_esnext/`, while `main` in package.json points to `.`

### DIFF
--- a/packages/cli-lib/package.json
+++ b/packages/cli-lib/package.json
@@ -53,7 +53,7 @@
     "translate",
     "translation"
   ],
-  "main": "index.js",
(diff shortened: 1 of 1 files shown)


pablobirukov:
@longlho any chance to review this? 

In 4.7.2 `"main"` points to non-existing `"index.js"` file, https://www.npmjs.com/package/@formatjs/cli-lib?activeTab=code while the built source is in `lib_esnext/`, so either build structure should be changed, or entry point updated to `"lib_esnext/index.js"`

PR #5032, opened 2025-06-23
fix(react-intl): add keys on nested rich text elements

zyzo:
Follow-up of https://github.com/formatjs/formatjs/pull/4895. When XML elements are nested `assignUniqueKeysToParts` uses the old hack `React.Children.toArray` to  generates keys, and React now logs an error message for those: 

```
Each child in a list should have a unique "key" prop.

Check the render method of `FormattedMessage`. See https://react.dev/link/warning-keys for more information.
```

Reproduction: 
https://codesandbox.io/p/sandbox/react-intl-issues-mr6v88

Capture d’écran 2025-06-23 à 10 30 30

This fix uses `toKeyedReactNodeArray` to properly silent the error message.

On a side note, the tests are still passing when I ran them against react 19. I guess it's because warning & error messages are permitted in tests. Would something like https://www.npmjs.com/package/jest-fail-on-console useful to be enabled?

### DIFF
--- a/packages/react-intl/src/utils.ts
+++ b/packages/react-intl/src/utils.ts
@@ -73,7 +73,7 @@ export function assignUniqueKeysToParts(
 ): FormatXMLElementFn<React.ReactNode> {
   return function (parts: any) {
     // eslint-disable-next-line prefer-rest-params
-    return formatXMLElementFn(React.Children.toArray(parts)) as any
+    return formatXMLElementFn(toKeyedReactNodeArray(parts)) as any
   }
 }
 


longlho (approved):

lucasbasquerotto:
Is there an ETA for when this fix will be added (a new version of react-intl)? For those using React 19, this issue is happening for months, even after the fix was implemented.

Haraldson:
Just upgraded to react@19.2.0 and react-intl@7.1.13, and key warnings started appearing in my console. Am I misunderstanding something, isn’t this supposed to be a fix for just that?

And just to be clear, I’m getting the same warnings with react-intl@7.1.12.

jozef-mikusinec-kiwicom:
The codesandbox still reproduces the error even if dependencies are updated to react@19.2.0 and react-intl@7.1.13

A workaround is to manually add a key prop to every value:

```
<FormattedMessage
  defaultMessage="<b><i>!</i></b>"
  id="key"
  values={{
    b: (chunks) => <b key="asdf">{chunks}</b>,
    i: (msg) => <i key="fofo">{msg}</i>,
  }}
/>
```

lucasbasquerotto:
> A workaround is to manually add a key prop to every value

We have dozens, maybe hundreds, of cases across different pages in our project with this issue. Making such a change would require a major refactor, while the real problem should be addressed at the source. Furthermore, the error logs do not indicate exactly where the `FormatMessage` component is being used, only one of its ancestors. This makes it time-consuming to find the actual instance causing the error, which would not be practical to fix for all cases. It is strange that the new version still does not resolve this issue. Is this PR included in version 7.1.13?

Haraldson:
@longlho Hi! Did you notice the feedback in these comments? I’m now at 7.1.14 and the issue persists.

lucasbasquerotto:
It seems fixed in the last version (`8.0.9`):

https://codesandbox.io/p/sandbox/react-intl-issues-forked-2z6gy8

PR #5035, opened 2025-06-24
Mention undocumented `preserveWhitespace` option in plugin docs

Romex91:
Provide some clarity for this issue: https://github.com/formatjs/formatjs/issues/935


Romex91:
@longlho @skoging the option was added here, but no docs:  https://github.com/formatjs/formatjs/pull/2592

longlho (approved):

PR #5046, opened 2025-06-30
fix: string list from iterable

septs:
```typescript
const formatter = new Intl.ListFormat()

function* getTexts() {
    yield "1"
    yield "2"
}

formatter.format(getTexts()) // Returns "1 and 2"
```

but polyfill not supported this behavior

--------

v8 implementation:

https://github.com/v8/v8/blob/3b25013317aaee1eeed5a513c4129d897b293b73/src/builtins/builtins-iterator-gen.cc#L283

### DIFF
--- a/packages/intl-listformat/index.ts
+++ b/packages/intl-listformat/index.ts
@@ -106,10 +106,7 @@ function validateInstance(instance: any, method: string) {
  * @param iterable list
  */
 function stringListFromIterable(iterable: Iterable<unknown>): string[] {
-  if (typeof iterable !== 'object' || iterable === null) return []
-  if (typeof iterable[Symbol.iterator] !== 'function') {
-    throw new TypeError('input not is valid iterable object')
-  }
+  if (typeof iterable !== 'object') return []
   const elements: string[] = []
   const iterator = iterable[Symbol.iterator]()
   let result: IteratorResult<unknown>
@@ -117,7 +114,9 @@ function stringListFromIterable(iterable: Iterable<unknown>): string[] {
     result = iterator.next()
     if (result.done) break
     if (typeof result.value !== 'string') {
-      throw new TypeError(`ArrayList[${elements.length}] is not type String`)
+      throw new TypeError(
+        `Iterable yielded ${elements.length} which is not a string`
+      )
     }
     elements.push(result.value)
   }
--- a/packages/intl-listformat/index.ts
+++ b/packages/intl-listformat/index.ts
@@ -114,9 +114,8 @@ function stringListFromIterable(iterable: Iterable<unknown>): string[] {
     result = iterator.next()
     if (result.done) break
     if (typeof result.value !== 'string') {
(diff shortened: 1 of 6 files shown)


packages/intl-listformat/index.ts
```diff
@@ -103,20 +103,24 @@ function validateInstance(instance: any, method: string) {
...
+  if (typeof iterable !== 'object') return []
+  const elements: string[] = []
+  const iterator = iterable[Symbol.iterator]()
+  let result: IteratorResult<unknown>
+  while (true) {
+    result = iterator.next()
+    if (result.done) break
+    if (typeof result.value !== 'string') {
+      throw new TypeError(
+        `Iterable yielded ${elements.length} which is not a string`
```
septs:
```typescript
// Uncaught TypeError: Iterable yielded 1 which is not a string
new Intl.ListFormat().format([1])
```

Test on Chrome 137.0.7151.120

packages/intl-listformat/index.ts
```diff
@@ -103,20 +103,24 @@ function validateInstance(instance: any, method: string) {
...
-    return []
-  }
-  const result = []
-  for (const el of list) {
-    if (typeof el !== 'string') {
-      throw new TypeError(`array list[${list.indexOf(el)}] is not type String`)
+function stringListFromIterable(iterable: Iterable<unknown>): string[] {
+  if (typeof iterable !== 'object') return []
+  const elements: string[] = []
+  const iterator = iterable[Symbol.iterator]()
```
septs, line 111:
```typescript
// Uncaught TypeError: object null is not iterable (cannot read property Symbol(Symbol.iterator))
new Intl.ListFormat().format(null)
```

packages/intl-listformat/index.ts
```diff
@@ -103,20 +103,24 @@ function validateInstance(instance: any, method: string) {
...
-function stringListFromIterable(list: any[]): string[] {
-  if (list === undefined) {
-    return []
-  }
-  const result = []
-  for (const el of list) {
-    if (typeof el !== 'string') {
-      throw new TypeError(`array list[${list.indexOf(el)}] is not type String`)
+function stringListFromIterable(iterable: Iterable<unknown>): string[] {
+  if (typeof iterable !== 'object') return []
```
septs, line 109:
```typescript
new Intl.ListFormat().format(undefined)
// Returns: ''
```

packages/intl-listformat/index.ts
septs: error message definition:

https://github.com/v8/v8/blob/3b25013317aaee1eeed5a513c4129d897b293b73/src/common/message-template.h#L137

v8 implementation:

https://github.com/v8/v8/blob/3b25013317aaee1eeed5a513c4129d897b293b73/src/builtins/builtins-iterator-gen.cc#L283

longlho (approved):

PR #5124, opened 2025-10-04
chore: Lerna-Lite --skip-bump-only-releases

ghiscoding:
- add `--skip-bump-only-releases` option to skip all the "bump only" changelog from the GitHub Release, this reduces the noise by skipping "bump only" from appearing in the GitHub Releases (though they still appear in their respective changelogs)
  - see https://github.com/lerna-lite/lerna-lite/tree/main/packages/version#--skip-bump-only-releases for more info
- also rename previous flag name `--remove-package-fields` to the new flag name `--strip-package-keys` because the old name will be removed in the next major version (it's a simple rename, the functionality is exactly the same as before)

The skip "bump only" releases is to avoid creating empty entries for something like below. You might still want to keep them, it's really up to you, this option is just a visual thing (it's useful if you want to only show the packages that really had changes and skip the "bump only" ones, as below)

image

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -9,13 +9,14 @@
       "exact": true,
       "ignoreChanges": ["**/tests/**", "**/*.md"],
       "message": "build: publish",
-      "private": false
+      "private": false,
+      "skipBumpOnlyReleases": true
     },
     "run": {
       "stream": true
     },
     "publish": {
-      "removePackageFields": ["devDependencies", "scripts"]
+      "stripPackageKeys": ["devDependencies", "scripts"]
     }
   },
   "packages": ["packages/*"]


longlho:
thanks for the PR. We have strict version dep, so I assume bump releases need to be created?

ghiscoding:
I think you misunderstood, it doesn't change anything about version or publish, all the versions will function the same as before. The only thing it does is to show less empty changes ("bump only") in the GitHub Release tab. So again, it's just a visual thing, it doesn't change anything about your versions and "bump only" versions will still be created, but they won't show in the Release tab with that flag

image

longlho:
Gotcha thanks for the clarification

longlho (approved):

PR #5753, opened 2026-01-01
docs: clarify ICU select, fix #4281

### LINKED ISSUE #4281: Documentation about the `{select} format` is unclear
## Which package?

Docs

Not knowing anything about the ICU Messsage syntax, I find [the following paragraph](https://formatjs.io/docs/core-concepts/icu-syntax#select-format) a bit unclear:

https://github.com/formatjs/formatjs/blob/58d6a7b398d776ca3d2726d72ae1573b65cc3bef/website/docs/core-concepts/icu-syntax.md?plain=1#L116-L124

In particular:

> The `matches` is a space-separated list of `matches`.

So `matches` is defined by itself? Is it how it should be or is it an error and `matches` is actually defined as a list of "match" structure, that would need to be properly defined?

> The format of a match is match `{output}`.

Having the block quotes only on _{output}_ is a bit confusing, I think it should also include _match_ just before.

> output is itself a message

I think it would be better if _output_ would be surrounded by block quotes as well?

> The `other` match is special and is used if nothing else matches.

What happens if `other` is already a value in the input data? I guess this case's already handled in the ICU spec but might be worth adding it here as well.

longlho:
### TL;DR

Improved the documentation for the ICU `select` format to be more precise and clear.

### What changed?

- Clarified the explanation of the `select` format by adding backticks around key terms for better readability
- Added more precise descriptions of how match cases work in the `select` format
- Added an important note explaining that `other` is treated as a reserved keyword, not as a variable lookup
- Improved the comparison to programming language constructs by adding backticks around terms like `case` and `default`

### How to test?

Review the updated documentation in the ICU syntax page to ensure it accurately describes the `select` format behavior, particularly the explanation of how the `other` keyword functions as a reserved fallback case.

### Why make this change?

The previous documentation was slightly ambiguous about how the `select` format works, particularly regarding the `other` keyword. This change provides more precise terminology and a clearer explanation of the matching process, which will help developers better understand how to use the `select` format correctly in their internationalization implementations.

### DIFF
--- a/docs/src/docs/core-concepts/icu-syntax.mdx
+++ b/docs/src/docs/core-concepts/icu-syntax.mdx
@@ -140,13 +140,13 @@ Similar to `number` type, we also support ICU DateTime skeleton. ICU provides a 
 
 ### `{select}` Format
 
-The `{key, select, matches}` is used to choose output by matching a value to one of many choices. (It is similar to the `switch` statement available in some programming languages.) The key is looked up in the input data. The corresponding value is matched to one of matches and the corresponding output is returned. The `key` argument must follow [Unicode Pattern_Syntax](https://www.unicode.org/reports/tr31/#Pattern_Syntax). The `matches` is a space-separated list of `matches`.
+The `{key, select, matches}` is used to choose output by matching a value to one of many choices. (It is similar to the `switch` statement available in some programming languages.) The `key` is looked up in the input data. The corresponding value is matched to one of the `matches` and the corresponding `output` is returned. The `key` argument must follow [Unicode Pattern_Syntax](https://www.unicode.org/reports/tr31/#Pattern_Syntax).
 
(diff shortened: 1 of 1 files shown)


longlho:
* **#5753** <a href="https://app.graphite.com/github/pr/formatjs/formatjs/5753?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/formatjs/formatjs/5753?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/formatjs?ref=merge-queue-instructions-in-stack-comment&prId=5753).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #5759, opened 2026-01-01
feat(eslint-plugin-formatjs): add allowWithDescription option to no-useless-message rule

lightpriest:
I opened this because it seemed pretty trivial to support and modify with backwards compatibility, and it seemed like other people could use.

Add a new allowWithDescription option to the no-useless-message lint rule that allows 'useless' messages when a description is provided. This provides an escape hatch for cases where the original language fits the replacement word exactly but the message ID still needs to exist so other locales can provide their own translations.

The default behavior is preserved when the option is not set or is false.

### DIFF
--- a/docs/src/docs/tooling/linter.mdx
+++ b/docs/src/docs/tooling/linter.mdx
@@ -822,6 +822,40 @@ This bans messages that do not require translation.
 
 Messages like `{test}` is not actionable by translators. The code should just directly reference `test`.
 
+#### Options
+
+```js
+
+export default [
+  {
+    plugins: {
+      formatjs,
+    },
+    rules: {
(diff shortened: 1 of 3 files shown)


longlho:
Hmm I don't understand the use case can u provide some examples?

lightpriest:
Yeah it's not as straight-forward like the default behavior. In our specific case we use formatjs to provide messages under different uses (which includes locales as well).

For example, in default locale you might want to use a value directly returned from the server (e.g. `{userNickname}`), and in a different locale you want to prefix it with something (e.g. `AKA {userNickname}`).

longlho:
hmm in ur case it's prob to just `// eslint-disable-next-line` tbh. I think this is pretty niche.

PR #5775, opened 2026-01-02
fix(@formatjs/ts-transformer): throws when message desc values are not statically analyzable, fix #4235

### LINKED ISSUE #4235: Catching Non Static Strings
In past implementations I've done I recall that when a FormattedMessage was not statically analyzable it would throw an error. The currently experience I'm seeing is that it just omits those messages from the output file silently. 

For example if I have the following in a component

```jsx
<FormattedMessage defaultMessage={`${Math.random()} not statically analyzable`} />
```

and then run the following

```
formatjs extract 'src/**/*.ts*' --ignore='**/*.d.ts' --out-file lang/en.json --id-interpolation-pattern '[sha512:contenthash:base64:6]' --throws
```

The task cli successfully runs with no additional output, but the output file is empty.

Is there a way I am missing somewhere to highlight the fact that there are invalid translations in the codebase? There doesn't seem to be an ESLint rule or way to make the extraction fail.

longlho:
## Fix: Extract non-static `defaultMessage` and `id` should throw error with `--throws` flag

Fixes #4235

### Problem

The formatjs CLI was silently succeeding when encountering non-static `defaultMessage` or `id` values instead of throwing an error when the `--throws` flag was used. This resulted in:
- Exit code 0 (success) instead of 1 (failure)
- Empty JSON output `{}` instead of an error message
- No indication that messages were being skipped due to non-static values

This was problematic because developers expect `--throws` to fail the build when messages can't be statically extracted, but the CLI was silently omitting them instead.

### Solution

Modified the TypeScript transformer to throw errors when encountering non-static values for `defaultMessage` or `id` fields. The fix adds validation at four strategic locations in the extraction pipeline:

1. After template literal evaluation (if non-static)
2. In JSX expression catch-all (for any unhandled expression type)
3. After binary expression evaluation (if non-static)  
4. In final catch-all (for any unhandled expression type)

The error messages differentiate between:
- `"must be a string literal or statically evaluable expression"` for binary expressions
- `"must be a string literal"` for other expression types

**Examples of non-static values that now throw errors:**
```tsx
// Template literal with expression
<FormattedMessage defaultMessage={`${dynamicValue} text`} />

// Variable concatenation
<FormattedMessage defaultMessage={`Value: ${variable}`} />

// Variable reference
<FormattedMessage defaultMessage={variable} />

// formatMessage with non-static string
intl.formatMessage({ defaultMessage: `Dynamic: ${value}` })

### DIFF
--- a/packages/cli/integration-tests/extract/integration.test.ts
+++ b/packages/cli/integration-tests/extract/integration.test.ts
@@ -256,6 +256,15 @@ test('invalid syntax should throw', async () => {
   }).rejects.toThrowError('TS1005')
 }, 20000)
 
+// https://github.com/formatjs/formatjs/issues/4235
+test('#4235: non-static defaultMessage should throw with --throws', async () => {
+  await expect(async () => {
+    await exec(
+      `${BIN_PATH} extract --throws '${join(__dirname, 'typescript/non-static.tsx')}'`
+    )
+  }).rejects.toThrowError(/defaultMessage.*must be a string literal/i)
+}, 20000)
+
 test('whitespace and newlines should be preserved', async () => {
   process.chdir(__dirname)
   await expect(
--- a/packages/cli/integration-tests/extract/typescript/non-static.tsx
+++ b/packages/cli/integration-tests/extract/typescript/non-static.tsx
@@ -1,0 +1,22 @@
+import React from 'react'
+import {FormattedMessage} from 'react-intl'
+
+// #4235: Non-static strings should throw an error when --throws is used
+export function NonStaticMessages() {
+  const dynamicValue = Math.random()
+
+  return (
+    <div>
+      {/* This should throw an error - template literal with expression */}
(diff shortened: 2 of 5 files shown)


longlho:
* **#5775** <a href="https://app.graphite.com/github/pr/formatjs/formatjs/5775?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/formatjs/formatjs/5775?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/formatjs?ref=merge-queue-instructions-in-stack-comment&prId=5775).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #5858, opened 2026-01-09
Fix import now that the package is ESM

adrienharnay:
The following error arises when building with Webpack:

```
ModuleNotFoundError: Module not found: Error: Can't resolve './IsValidDateTimeFieldCode' in '/fakepath/node_modules/@formatjs/intl-displaynames/abstract'
Did you mean 'IsValidDateTimeFieldCode.js'?
BREAKING CHANGE: The request './IsValidDateTimeFieldCode' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

Adding the extension fixes the error.

### DIFF
--- a/packages/intl-displaynames/abstract/CanonicalCodeForDisplayNames.ts
+++ b/packages/intl-displaynames/abstract/CanonicalCodeForDisplayNames.ts
@@ -4,7 +4,7 @@ import {
   IsWellFormedCurrencyCode,
 } from '@formatjs/ecma402-abstract'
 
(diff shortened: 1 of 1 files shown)


longlho:
Thanks a lot for your contributions! fixed in latest

PR #5895, opened 2026-01-14
docs: make dark colour scheme default

samrobbins85:
For admonitions and search they're set up with the tailwind dark colour scheme, but the site doesn't implement that, instead being all dark, making the admonitions difficult to read e.g. under this section https://formatjs.github.io/docs/react-intl#the-react-intl-module

Before:
Screenshot 2026-01-14 at 16 13 01

After:
Screenshot 2026-01-14 at 16 13 22


longlho:

longlho (approved):

PR #6002, opened 2026-01-31
fix(@formatjs/cli-lib): respect throws:false flag in extract() function

### LINKED ISSUE #5994: --throws flag seemingly not respected with extract() function
## Which package?

ts-transform

## Describe the bug

This occurred when I upgraded to 8.1.0 and beyond of @formatjs/cli-lib.
We use extract() in our tests to verify that all our intl strings actually exists in the code.

```
const foundTranslations = await extract(files, {
        idInterpolationPattern: '[sha512:contenthash:base64:6]'
});
```
 Prior to 8.1.0 we were relying on erroneous keys simply being skipped. But after this it throws as soon as an erroneous key is found.

I think I have narrowed it down to this commit here: https://github.com/formatjs/formatjs/commit/16fdf64951e5b9f09e65d6ee92507cf3eab456bd#diff-25baed40cc703386cb8cbece67f207cf31ee23a36ff94c0ffead2f32aede87d2

The commit is well documented and it seems to me I can revert to the old behaviour by adding throws: false.
The comment here in this test alludes to that: https://github.com/formatjs/formatjs/commit/16fdf64951e5b9f09e65d6ee92507cf3eab456bd#diff-9c026660716a6d9efef0b08c7288f031230963af8edb13184b7ca2e8c013ef18R4

```
const foundTranslations = await extract(files, {
        throws: false, <-- I thought adding this would resume previous behavior
});
```

However, what I think happens is that though the throws flag is handled here
https://github.com/formatjs/formatjs/blob/16fdf64951e5b9f09e65d6ee92507cf3eab456bd/packages/cli-lib/src/extract.ts#L220

The promise.all is still rejected because an error was thrown, resulting in rawResults being empty evne though several good messages was extracted.
https://github.com/formatjs/formatjs/blob/16fdf64951e5b9f09e65d6ee92507cf3eab456bd/packages/cli-lib/src/extract.ts#L211

First time looking at this repository so apologise if I have misunderstood anything.

I don't know what is the desired behaviour here from your side. But I don't think there is any way I can work around this change from our part 🤔 We are relying on getting what could be extracted, and handling those messages who did not ourselves.

### Codesandbox URL

Will provide if required

## Expected behavior

I expect that setting `throws: false` will allow me to use extract as before that stricter id interpolation checks were added. We use sporadic dynamic keys, so I cannot rewrite to always using string literals for extraction

- OS: iOS, linux
- Browser: JSdom

longlho:
When throws is set to false, the extract() function should collect partial results from all files instead of failing completely when one file has an error. This fix uses Promise.allSettled() when throws is false to handle individual file failures gracefully.

Fixes #5994

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/formatjs_cli/src/extract.rs
+++ b/crates/formatjs_cli/src/extract.rs
@@ -1390,4 +1390,139 @@ const msg = defineMessage({
             "Extracting the same file twice should produce identical output"
         );
     }
+
+    #[test]
+    fn test_extract_with_throws_false_collects_partial_results() {
+        // Test that when throws:false, extraction collects partial results from valid files
+        // even when some files have errors
+        let temp_dir = tempfile::tempdir().unwrap();
+        let output_file = temp_dir.path().join("output.json");
+
+        // Create a valid file
+        let valid_file = temp_dir.path().join("valid.tsx");
+        let valid_content = r#"
+import { defineMessage } from 'react-intl';
+
+const msg = defineMessage({
+  id: 'valid.message',
+  defaultMessage: 'This is a valid message',
+});
+"#;
+        std::fs::write(&valid_file, valid_content).unwrap();
+
+        // Create an invalid file (malformed JavaScript)
+        let invalid_file = temp_dir.path().join("invalid.tsx");
+        let invalid_content = r#"
+import { defineMessage } from 'react-intl';
+
+const msg = defineMessage({
+  id: 'invalid.message
+  defaultMessage: 'This is missing closing quote and brace
+"#;
+        std::fs::write(&invalid_file, invalid_content).unwrap();
+
+        // Create another valid file
+        let valid_file2 = temp_dir.path().join("valid2.tsx");
+        let valid_content2 = r#"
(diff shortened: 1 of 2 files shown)


longlho:
@claude verify our Rust impl as well

longlho:
@claude fix the commit message so our hook passed

longlho:
* **#6002** <a href="https://app.graphite.com/github/pr/formatjs/formatjs/6002?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/formatjs/formatjs/6002?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/formatjs?ref=merge-queue-instructions-in-stack-comment&prId=6002).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

PR #6044, opened 2026-02-25
fix(@formatjs/vite-plugin): wrap JSX defaultMessage AST in expression container

### LINKED ISSUE #6042: error in vite-plugin
## Which package?

vite-plugin

## Describe the bug

When setting `ast: true` in vite.config.ts, the application crashes during development with the following error:
`Pre-transform error: index.tsx: JSX value should be either an expression or a quoted JSX text`

The issue only occurs when ast is enabled. The same code works correctly when ast is set to false.

The JSX that triggers the error looks like this:
`<FormattedMessage id="PRVGwS" defaultMessage=[{"type":0,"value":"Hello World"}] />`

## To Reproduce

- Create a simple Vite + React project.
- Add the vite-plugin to vite.config.ts.
- Enable ast: true in the plugin configuration.
- Add some string in the project with  `<FormattedMessage defaultMessage="Hello World" />`
- Run the development server.

## Expected behavior
The application should run normally with `ast: true` enabled, without throwing a JSX transformation error.

## Additional context

I have babel-plugin-react-compiler, nodePolyfills and tailwindcss plugins wired

Adding the plugin like this in my vite.config.ts
` formatjs({
      idInterpolationPattern: "[sha512:contenthash:base64:6]",
      ast: true,
    })`

longlho:
When `ast: true` is enabled in the vite-plugin, the `defaultMessage` prop in JSX elements was being replaced with a bare JSON array, producing invalid JSX:

```jsx
<FormattedMessage defaultMessage=[{"type":0,"value":"Hello World"}] />
```

The fix wraps the serialized AST in a JSX expression container `{...}` when in JSX context:

```jsx
<FormattedMessage defaultMessage={[{"type":0,"value":"Hello World"}]} />
```

Fixes #6042

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/vite-plugin/tests/transform.test.ts
+++ b/packages/vite-plugin/tests/transform.test.ts
@@ -111,11 +111,11 @@ describe('@formatjs/vite-plugin transform', () => {
     })
 
     test('wraps JSX defaultMessage AST in expression container', () => {
-      const input = `<FormattedMessage defaultMessage="Hello World" />`
+      const input = `<FormattedMessage id="test" defaultMessage="Hello World" />`
       const output = t(input, {ast: true})
-      // Should use JSX expression container syntax {[...]} not bare [...]
-      expect(output).toContain('defaultMessage={[')
-      expect(output).toContain('"type"')
+      expect(output).toBe(
+        `<FormattedMessage id="test" defaultMessage={[{"type":0,"value":"Hello World"}]} />`
+      )
     })
   })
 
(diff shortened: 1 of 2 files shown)


packages/vite-plugin/tests/transform.test.ts
```diff
@@ -109,6 +109,14 @@ describe('@formatjs/vite-plugin transform', () => {
       expect(output).toContain('"type"')
       expect(output).not.toContain('description')
     })
+
+    test('wraps JSX defaultMessage AST in expression container', () => {
+      const input = `<FormattedMessage defaultMessage="Hello World" />`
+      const output = t(input, {ast: true})
+      // Should use JSX expression container syntax {[...]} not bare [...]
+      expect(output).toContain('defaultMessage={[')
```
longlho: @claude don't check for contain, check the whole output

PR #6065, opened 2026-03-09
feat(eslint-plugin-formatjs): add prefer-direction-agnostic stylesheet rule

cungminh2710:
Adds a new prefer-direction-agnostic ESLint rule to help catch physical CSS directional styles that break RTL layouts.

This rule:

- checks .css, .scss, .sass, and .less files
- flags properties like margin-left, padding-right, left, right, and directional border/corner styles
- flags values like text-align: left/right, float: left/right, and clear: left/right
- provides autofixes to logical equivalents such as margin-inline-start, inset-inline-end, and text-align: start/end

More broadly, this moves us toward keeping i18n-related linting in eslint-plugin-formatjs, including stylesheet rules that directly affect localization and RTL support.

### DIFF
--- a/docs/src/docs/tooling/linter.mdx
+++ b/docs/src/docs/tooling/linter.mdx
@@ -1,4 +1,4 @@
-This eslint plugin allows you to enforce certain rules in your ICU message.
+This eslint plugin allows you to enforce rules for ICU messages and related localization patterns in your source files.
 
 ## Usage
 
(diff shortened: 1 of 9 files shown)


longlho:
Thanks but we're not gonna take in a css parser as a dep, let alone postcss. Maybe make a stylelint plugin package?

PR #6066, opened 2026-03-10
feat(eslint-plugin-formatjs): support minimum description length in `enforce-description` rule

William-Feng:
## Problem

The `enforce-description` ESLint rule checks that a `description` field is present on message descriptors, but it doesn't enforce any quality bar. In practice, this means developers can satisfy the rule with descriptions like "Title" or "Button text" (or even a singular character like "x"), but such descriptions provide almost no useful context or value to translators.

## Summary

Adds an optional `minLength` option to the `enforce-description` ESLint rule so teams can enforce a minimum character length for message descriptions at lint time.

The rule now accepts either the existing string option or a new object form, maintaining full backward compatibility:
```js
// Existing - still works
'formatjs/enforce-description': ['error', 'literal']

// New object form - minLength and message are both optional
'formatjs/enforce-description': ['error', { minLength: 10, message: 'literal' }]
```

When a description is present but shorter than `minLength`, the rule reports:
> description must be at least 10 characters long (currently 8)

If the description is missing entirely, the existing `enforceDescription` error is reported as before.

### Example configuration

```js
// eslint.config.js
export default [
  {
    plugins: { formatjs },
    rules: {
      'formatjs/enforce-description': ['error', { minLength: 10, message: 'literal' }],
    },
  },
]
```

### Example lint output

❌ Fails
```js
defineMessage({
  defaultMessage: 'Hello',
  description: 'greet',   // `description` must be at least 10 characters long (currently 5)
})
```

✅ Passes
```js
defineMessage({
  defaultMessage: 'Hello',
  description: 'Greeting shown on the homepage banner',
})
```

 ### Test plan
 
[x] Added 8 test cases (3 valid, 5 invalid) across `defineMessage`, `intl.formatMessage` and `<FormattedMessage/>`
[x] All unit tests pass (`bazel test //packages/eslint-plugin-formatjs:unit_test`)

### DIFF
--- a/docs/src/docs/tooling/linter.mdx
+++ b/docs/src/docs/tooling/linter.mdx
@@ -197,6 +197,10 @@ const messages = defineMessages({
 
 #### Options
 
+The rule accepts a single option, which can be either a string or an object.
+
+**String option:**
+
 ```js
 export default [
   {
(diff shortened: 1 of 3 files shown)


longlho (approved):

longlho:

PR #6180, opened 2026-03-25
fix(deps): remove typescript peer dependencies

### LINKED ISSUE #6179: react-intl@7 with typescript@6 (possibly 7 in future)
Hello,

Yesterday, TypeScript 6 was released (with TypeScript 7 expected to follow within a few months, according to the announcement: https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/
).

I’m writing this as a feature request.

At the moment, many applications cannot easily migrate to React 19 due to peer dependency constraints. In some cases, these are just narrowly defined peer dependency ranges, while in others there may be actual compatibility issues. As a result, upgrading to newer versions of `react-intl` (v8+, currently v10) is not feasible for a lot of projects.

From what I understand, one of the reasons react-intl currently requires TypeScript 5 is related to types such as Intl.DisplayNames and Intl.ListFormat. However, these don’t appear to be mentioned as changed or impacted in the TypeScript 6 release. If that understanding is correct, it suggests that compatibility with TypeScript 6 (and possibly 7) might not require any significant changes.

Given this situation, would it be possible to release a `react-intl@7` version compatible with `typescript@6` (and potentially TypeScript 7 as well)? Since TypeScript 6 is intended as a bridge between 5.9 and 7, it likely won’t introduce additional breaking changes beyond what’s coming in 7.

I believe such a release would help many teams continue upgrading their TypeScript setup without being blocked by React version compatibility.

Thank you for considering this request.

camsteffen:
Fixes #6179

To my understanding, it's not typical to have typescript as a library peer dependency. Any typescript version compatibility issues will arise at build time. Peer dependencies are more valuable when ensuring runtime compatibility issues.

### DIFF
--- a/packages/intl/package.json
+++ b/packages/intl/package.json
@@ -16,9 +16,6 @@
     "@formatjs/icu-messageformat-parser": "workspace:*",
     "intl-messageformat": "workspace:*"
   },
-  "peerDependencies": {
-    "typescript": "^5.6.0"
-  },
   "bugs": "https://github.com/formatjs/formatjs/issues",
   "homepage": "https://formatjs.github.io",
   "keywords": [
@@ -34,10 +31,5 @@
     "translate",
     "translation"
   ],
-  "peerDependenciesMeta": {
-    "typescript": {
-      "optional": true
-    }
-  },
   "repository": "git@github.com:formatjs/formatjs.git"
 }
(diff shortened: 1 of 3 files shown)


longlho:
@claude fix this build

camsteffen:
Silly me - had a JSON syntax error

PR #6210, opened 2026-03-30
chore(@formatjs/intl-locale): bundle with esbuild, externalize deps

longlho:
## Summary
- Bundle intl-locale entry points with esbuild using `entry_points` + `output_dir`
- `@formatjs/*` and `cldr-*` deps marked as `external` — they stay as import statements in the output
- Type declarations still generated by `ts_compile`
- This ensures internal import paths (like `#packages/*`) never leak into the published npm package

The bundled JS overrides the ts_compile JS in the npm_package, while `.d.ts` files are preserved.

## Test plan
- [x] All 495 Bazel tests pass
- [x] Published `index.js` only contains `@formatjs/*` external imports, no internal paths
- [x] Pre-commit hooks pass

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -12,6 +12,13 @@ load("//tools:tsconfig.bzl", "BASE_TSCONFIG")
 # gazelle:js_pnpm_lockfile pnpm-lock.yaml
 # gazelle:js_validate_import_statements off
 
+# Exported for Node subpath import resolution (#packages/*) in tests.
+# Node walks up to find package.json with "imports" field.
+exports_files(
+    ["package.json"],
+    visibility = ["//visibility:public"],
+)
+
 gazelle(
     name = "gazelle",
     env = {
--- a/packages/intl-locale/BUILD.bazel
+++ b/packages/intl-locale/BUILD.bazel
@@ -1,15 +1,104 @@
 load("@aspect_bazel_lib//lib:copy_to_bin.bzl", "copy_to_bin")
 load("@aspect_bazel_lib//lib:write_source_files.bzl", "write_source_files")
 load("@aspect_rules_esbuild//esbuild:defs.bzl", "esbuild")
+load("@aspect_rules_js//js:defs.bzl", "js_library")
 load("@aspect_rules_js//npm:defs.bzl", "npm_package")
+load("@aspect_rules_ts//ts:defs.bzl", "ts_project")
+load("@npm//:@typescript/native-preview/package_json.bzl", tsgo_bin = "bin")
 load("@npm//:defs.bzl", "npm_link_all_packages")
 load("@npm//:test262-harness/package_json.bzl", test262_harness_bin = "bin")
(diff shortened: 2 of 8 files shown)


packages/intl-locale/BUILD.bazel
```diff
@@ -1,61 +1,129 @@
...
+BUNDLE_DEPS = [
+    ":node_modules/@formatjs/intl-supportedvaluesof",
+    ":node_modules/@formatjs/intl-getcanonicallocales",
+    "//:node_modules/cldr-core",
+    "//:node_modules/cldr-bcp47",
+    "//packages/ecma402-abstract",
+]
+
+# External packages preserved as imports in bundle output
+EXTERNAL_PACKAGES = [
```
longlho, line 52: derive this from BUNDLE_DEPS, exclude all non-`//:node_modules/`

packages/intl-locale/index.ts
```diff
@@ -5,7 +5,7 @@ import {
   SameValue,
   createDataProperty,
   invariant,
-} from '@formatjs/ecma402-abstract'
+} from '#packages/ecma402-abstract/index.ts'
```
longlho: why `.ts` should be `.js`?

tools/tsconfig.bzl
```diff
@@ -72,6 +72,13 @@ DOCS_TSCONFIG = BASE_TSCONFIG | {
     },
 }
 
+# Declaration-only configuration - emits .d.ts but no .js (for packages bundled with esbuild)
+DECL_ONLY_TSCONFIG = BASE_TSCONFIG | {
```
longlho: this looks unused

BUILD.bazel
```diff
@@ -12,6 +12,11 @@ load("//tools:tsconfig.bzl", "BASE_TSCONFIG")
 # gazelle:js_pnpm_lockfile pnpm-lock.yaml
 # gazelle:js_validate_import_statements off
 
+exports_files(
+    ["package.json"],
```
longlho, line 18: we technically don't need the whole packagejson, we just need a few fields like `type: modules` and `imports` right?

packages/intl-locale/BUILD.bazel
longlho: You're right, we only need `type` and `imports`. But Node walks up the directory tree looking for a `package.json` — the file must be named `package.json` at the repo root. Generating a minimal one with a genrule would conflict with the real `package.json`. Since the real one already has the fields and it's a source file (not a heavy build artifact), I think using it directly is the simplest approach. Happy to change if you have a different idea.

packages/intl-locale/index.ts
longlho: Fixed — changed to `.js`. TypeScript with `rewriteRelativeImportExtensions` + `moduleResolution: bundler` resolves `.js` imports to the `.ts` source during type checking. And esbuild's plugin does the `.js` → `.ts` fallback at bundle time.

PR #6213, opened 2026-04-01
fix(icu-messageformat-parser): use Map.has() instead of in operator in collectVariables

### LINKED ISSUE #6212: @formatjs/icu-messageformat-parser falsely reports conflicting types for variables like size and entries
## Which package?

`@formatjs/icu-messageformat-parser`

## Describe the bug

`isStructurallySame` appears to throw a false variable type conflict when an ICU variable name collides with a property or method name on `Map`, such as `size` or `entries`.

For example, structurally identical messages like `{size} m²` and `{date} · {size}` fail with:

- `Variable size has conflicting types`
- `Variable entries has conflicting types`

This looks like a false positive rather than a real ICU type mismatch.

### Codesandbox URL

N/A

### Reproducible Steps/Repo

1. Import `parse` and `isStructurallySame` from `@formatjs/icu-messageformat-parser`
2. Run the following code:

```ts
import {parse, isStructurallySame} from '@formatjs/icu-messageformat-parser'

console.log(isStructurallySame(parse('{size} m²'), parse('{size} m²')))
console.log(isStructurallySame(parse('{date} · {size}'), parse('{date} · {size}')))
console.log(
  isStructurallySame(
    parse("Congratulations! You’ve won {entries} entries."),
    parse("Congratulations! You’ve won {entries} entries.")
  )
)
```

3. Observe that it throws or reports:
   - `Variable size has conflicting types`
   - `Variable entries has conflicting types`

## Expected behavior

`isStructurallySame(...)` should succeed for these examples because the source and target messages are structurally identical.

## Screenshots

N/A

- OS: macOS
- Browser: N/A
- Version: N/A

- Device: N/A
- OS: N/A
- Browser: N/A
- Version: N/A

## Additional context

We traced this to `collectVariables` in `packages/icu-messageformat-parser/manipulator.ts`:

```ts
if (el.value in vars && vars.get(el.value) !== el.type) {
  throw new Error(`Variable ${el.value} has conflicting types`)
}
```

`vars` is a `Map`, and this check seems unsafe for variable names like `size` and `entries`.

It looks like this should use `vars.has(el.value)` instead:

```ts
if (vars.has(el.value) && vars.get(el.value) !== el.type) {
  throw new Error(`Variable ${el.value} has conflicting types`)
}
```

We observed this with:

- `@formatjs/icu-messageformat-parser` `3.5.3`

Gdhanush-13:
### What

Fixes #6212

### Problem

In `collectVariables()`, the check `el.value in vars` uses the `in` operator on a `Map` object. The `in` operator checks the **entire prototype chain**, so ICU variable names like `size`, `entries`, `get`, `set`, `has`, `delete`, `clear`, `forEach`, and `keys` falsely match `Map.prototype` properties. This causes a spurious `Variable {name} has conflicting types` error even when no conflict exists.

### Root Cause

```ts
// Before (buggy): 'in' checks prototype chain
if (el.value in vars && vars.get(el.value) !== el.type) {
```

For example, `'size' in new Map()` returns `true` because `Map.prototype.size` exists as a getter.

### Fix

```ts
// After: .has() only checks the Map's own entries
if (vars.has(el.value) && vars.get(el.value) !== el.type) {
```

Also added a regression test covering variable names that collide with `Map` prototype properties.

### DIFF
--- a/packages/icu-messageformat-parser/manipulator.ts
+++ b/packages/icu-messageformat-parser/manipulator.ts
@@ -172,6 +172,9 @@ function collectVariables(
       isTimeElement(el) ||
       isNumberElement(el)
     ) {
+      if (vars.has(el.value) && vars.get(el.value) !== el.type) {
+        throw new Error(`Variable ${el.value} has conflicting types`)
+      }
       vars.set(el.value, el.type)
     }
 
--- a/packages/icu-messageformat-parser/manipulator.ts
+++ b/packages/icu-messageformat-parser/manipulator.ts
@@ -172,9 +172,6 @@ function collectVariables(
       isTimeElement(el) ||
       isNumberElement(el)
     ) {
-      if (el.value in vars && vars.get(el.value) !== el.type) {
-        throw new Error(`Variable ${el.value} has conflicting types`)
-      }
       vars.set(el.value, el.type)
     }
 
--- a/packages/icu-messageformat-parser/tests/manipulator.test.ts
+++ b/packages/icu-messageformat-parser/tests/manipulator.test.ts
@@ -112,6 +112,34 @@ test('structurally same literal', function () {
   )
 })
 
+// https://github.com/formatjs/formatjs/issues/6212
+test('GH #6212: variable names matching Map prototype properties should not conflict', function () {
+  // 'size' and 'entries' are properties on Map.prototype.
+  // Using `in` operator instead of `Map.has()` caused false positives.
+  expect(
+    isStructurallySame(parse('{size} m²'), parse('{date} · {size}'))
+  ).toEqual({
+    success: false,
+    error: new Error(
+      'Different number of variables: [size] vs [date, size]'
+    ),
+  })
+
+  // Same variables — should succeed without "conflicting types" error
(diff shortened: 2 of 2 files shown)


longlho:

packages/icu-messageformat-parser/manipulator.ts
```diff
@@ -172,9 +172,6 @@ function collectVariables(
       isTimeElement(el) ||
       isNumberElement(el)
     ) {
-      if (el.value in vars && vars.get(el.value) !== el.type) {
```
longlho, line 175: hmm this doesn't look right. It shouldn't be deleted right?

longlho (changes requested):

packages/icu-messageformat-parser/manipulator.ts
Gdhanush-13: You're right, sorry about that! I've restored the type conflict detection - now using `vars.has(el.value)` instead of the `in` operator, which was the actual root cause of the bug. The conflict check is back in place.

PR #6391, opened 2026-04-18
build: add @formatjs_generated package infrastructure

longlho:
## Summary
- Add `tools/generated.bzl` with `generate_package_file()` and `formatjs_generated_package()` macros for creating Bazel-only generated npm packages
- Add `tools/generated_packages.bzl` with package registry and `link_all_generated_packages()` helper
- Update `tools/compile.bzl` to exclude `@formatjs_generated` from rolldown externals (generated data should be bundled inline)
- Update `tools/tsconfig.bzl` and `tools/index.bzl` with `@formatjs_generated/*` path mappings for IDE support
- Update `tools/gazelle/ts/resolve.go` to resolve `@formatjs_generated/` imports to Bazel labels
- Call `link_all_generated_packages()` from root `BUILD.bazel`

Infrastructure only — the package registry is empty, so there is no behavior change.

## Test plan
- [x] `bazel build //packages/intl-locale:pkg` succeeds
- [x] `bazel build //tools/gazelle/ts:ts` compiles
- [x] `pnpm t` passes


tools/generated.bzl
```diff
@@ -0,0 +1,93 @@
...
+        ],
+        chdir = chdir,
+        srcs = data + [
+            "//:node_modules/fs-extra",
+            "//:node_modules/minimist",
+        ],
+    )
+
+    # Format with oxfmt
+    native.genrule(
```
longlho: this is generated file we don't care about formatting

tools/generated_packages.bzl
```diff
@@ -0,0 +1,31 @@
+"Registry of all @formatjs_generated packages and linking helper."
+
+load("@aspect_rules_js//npm:defs.bzl", "npm_link_package")
+
+# All @formatjs_generated packages. Add new entries here when creating a new
+# generated package. The "name" is the npm package suffix under the
+# @formatjs_generated scope; "target" is the Bazel label of the npm_package.
+GENERATED_PACKAGES = [
```
longlho: separate this out to a bzl file, then write a script to generate this by using bazel query

PR #6423, opened 2026-04-21
build: migrate Rust toolchain from rules_rust to rules_rs

dzbarsky:
## Summary
- Migrate Rust build rules from `rules_rust` to `rules_rs` for improved Bazel module compatibility
- Update all Rust crate BUILD files to use `rules_rs` load paths
- Add custom platform definitions for Linux GNU and Windows MSVC
- Patch `rules_rust_wasm_bindgen` for rules_rs compatibility

## Known Issues (need fixing)
1. **`@llvm` not visible**: `platforms/BUILD.bazel` references `@llvm//constraints/libc:gnu.2.28` but `llvm` is only a transitive dep of `rules_rs`. Fix: add `bazel_dep(name = "llvm", version = "0.7.0")` to MODULE.bazel
2. **`ts_import_extractor` prost dep**: The old `@rules_rust_prost//private/3rdparty/crates:prost` path doesn't exist in rules_rs. The code in `main.rs` uses `prost::Message` directly, so prost must be a dep — but it needs to come from the same crate universe as the prost toolchain to avoid trait mismatch. Options:
   - Configure `crate.annotation` in MODULE.bazel to alias the toolchain's prost
   - Or add prost to Cargo.toml and configure `crate.from_cargo` with `use_experimental_platforms` to unify versions

## Test plan
- [ ] `bazel build //platforms:all` passes
- [ ] `bazel build //crates/...` passes
- [ ] `bazel test //packages/...` passes
- [ ] `bazel test //crates/...` passes


longlho (commented):
Two issues blocking CI:

### 1. `@llvm` not visible from root module
`platforms/BUILD.bazel` references `@llvm//constraints/libc:gnu.2.28` but `llvm` is a transitive dep of `rules_rs`, not directly visible. Fix:

```starlark
# MODULE.bazel, after rules_rs
bazel_dep(name = "llvm", version = "0.7.0")
```

### 2. `ts_import_extractor` prost runtime dep
The old `@rules_rust_prost//private/3rdparty/crates:prost` path doesn't exist in rules_rs. The code in `main.rs` uses `prost::Message` directly, so prost must be an explicit dep.

However, adding prost to `Cargo.toml` + `@crates//:prost` creates a **separate crate instance** from the one used by `rust_prost_library` (the toolchain uses `rrprd__prost-0.13.1`). Even at the same version, Rust treats them as different types, causing `Message` trait mismatch.

The fix needs `rules_rs`-specific config to either:
- Expose the toolchain's prost via `crate.annotation` 
- Or configure `crate.from_cargo` to share the prost instance with the toolchain

PR #6433, opened 2026-04-24
fix(@formatjs/ts-transformer): restore ts-jest-integration import

azolotopup:
PR #6080 added `"exports": { ".": "./index.js" }` to `@formatjs/ts-transformer`, which inadvertently broke the ts-jest-integration deep import that ts-jest users rely on for AST transformer configuration.

#### Problem
After upgrading to `@formatjs/ts-transformer@4.4.3`, any project using `ts-jest` with the formatjs ast transformer fails before tests execute with error:
`File not found: @formatjs/ts-transformer/ts-jest-integration`

This affects the documented ts-jest configuration pattern:

#### Fix
Expose `ts-jest-integration` as an explicit subpath export

#### Impact
No breaking changes — existing "." export is unchanged
Restores the documented ts-jest integration path without requiring any workaround on the consumer side
ts-jest-integration.js already exists in the published package; this change only makes it reachable via the exports map
#### Related
Regression introduced by: #6080

### DIFF
--- a/packages/ts-transformer/package.json
+++ b/packages/ts-transformer/package.json
@@ -7,7 +7,8 @@
   "type": "module",
   "types": "index.d.ts",
   "exports": {
-    ".": "./index.js"
+    ".": "./index.js",
+    "./ts-jest-integration.js": "./ts-jest-integration.js"
   },
   "engines": {
     "node": ">= 20.12.0"


packages/ts-transformer/package.json
```diff
@@ -7,7 +7,8 @@
   "type": "module",
   "types": "index.d.ts",
   "exports": {
-    ".": "./index.js"
+    ".": "./index.js",
+    "./ts-jest-integration": "./ts-jest-integration.js"
```
longlho: Let's do .js for esm consistency

longlho (changes requested):

longlho:
Thanks for the report! I've opened #6437 with a more complete fix that also adds `ts-jest-integration.ts` as a rolldown entry point (so the `.js`, `.d.ts`, and `.js.map` are actually bundled into the published package). Closing this in favor of that PR.

Issue #6435, opened 2026-04-24
ts-jest-integration is missing in @formatjs/ts-transformer 4.4.4


joshunger:
The file `ts-jest-integration.js` disappeared in 4.4.4.

https://www.npmjs.com/package/@formatjs/ts-transformer?activeTab=code

https://www.npmjs.com/package/@formatjs/ts-transformer/v/4.4.3?activeTab=code

Is there a recommended path for 4.4.4? Thanks!

longlho:
fix in https://github.com/formatjs/formatjs/pull/6433

joshunger:
Thank you @longlho!

PR #6437, opened 2026-04-24
fix(@formatjs/ts-transformer): restore ts-jest-integration subpath export

longlho:
## Summary
- Add `./ts-jest-integration.js` to the `exports` field in `package.json`
- Add `ts-jest-integration.ts` as a rolldown entry point in `BUILD.bazel` so it gets bundled with `.js`, `.d.ts`, and `.js.map` into the published package
- Update jest integration test config to use `.js` extension in the import path

This was a regression from #6080 which added an `exports` field to `package.json` but only included `.` — blocking the `ts-jest-integration` deep import that ts-jest users rely on.

Closes #6433

## Test plan
- [x] `bazel test //packages/ts-transformer:package_exports_test` — validates all exports resolve to real files
- [x] `bazel test //packages/ts-transformer:unit_test` — 27 tests pass
- [x] All pre-commit hooks pass


longlho:
fix #6435

PR #6438, opened 2026-04-26
build(deps): migrate gazelle TS plugin to hermeticbuild/gazelle_ts

longlho:
## Summary

- Replace the in-tree gazelle TypeScript extension (`tools/gazelle/ts/`) and the Rust oxc import-extractor (`crates/ts_import_extractor/`) with the upstream [`gazelle_ts`](https://github.com/hermeticbuild/gazelle_ts) module (formerly `gazelle_plugins`), pinned at v0.2.0.
- `# gazelle:map_kind` rewrites the stock `ts_project` / `js_test` rules to `formatjs_library` / `formatjs_test`. Macros now consume the plugin's flat `deps` / `data` shape and partition npm vs internal labels internally.
- Drops ~2.5k lines of vendored Go + Rust subprocess code in favor of a maintained upstream plugin.
- Removes the unused wasm cross-compile path from `crates/icu_messageformat_parser` (no consumer of the wasm artifact in this repo) — simplifies the Rust toolchain story so this PR doesn't need any LLVM cross-toolchain plumbing.

## Notable changes

- **`MODULE.bazel`** —
  - Adds `gazelle_ts` v0.2.0 via `archive_override` (BCR doesn't carry the renamed module yet). v0.2.0 ships dev_dependency-only toolchain + Cargo extensions, so its `MODULE.bazel` doesn't leak into consumers; no patch needed on our side.
  - Adds `import_extractor` module extension which downloads the prebuilt Rust binary that backs the plugin — the consumer doesn't build it from source.
  - Drops `bazel_dep(name = "rules_rust_wasm_bindgen")` + the `third_party/rules_rust_wasm_bindgen/use_rules_rs.patch`, and drops `wasm32-unknown-unknown` from `crate.from_cargo`'s `platform_triples`. Stays on `register_toolchains(\"@default_rust_toolchains//:all\")` only — no need for `@llvm//toolchain:all` once wasm is out (LLVM's clang++/lld can't link our native Linux artifacts).

- **`tools/compile.bzl`** — `formatjs_library` accepts a single combined `deps` and auto-partitions npm vs internal; emits an alias from `name` (e.g. `:dist`) to `:lib` for cross-package refs; forwards to `ts_project` when called with `no_emit = True` (handles `map_kind`'d manual `ts_project` rules); silently absorbs the `composite/declaration/declaration_map/source_map` kwargs gazelle emits.

- **`tools/test.bzl`** — `formatjs_test` consumes gazelle's `data` + `entry_point`, partitions into vitest srcs / deps / runtime data, and falls through to stock `js_test` for hand-rolled entry-point tests. Substitutes `:lib` (source-only `js_library`) for the plugin-emitted sibling-library label so the bundled `:basename` `package.json` doesn't shadow the root `//:package.json` and break Node's `#packages/*` resolution. New `extra_data` / `extra_srcs` kwargs cover hand-maintained list comprehensions gazelle's flat-list merger can't process (e.g. generated locale-data fixtures).

- **`tools/vitest.bzl`** — dedupes the final `data` list (`//:package.json` was getting added twice).

- **`tools/index.bzl`** — `generate_ide_tsconfig_json` strips `:target` suffix from project_references before computing the relative tsconfig path.

- **Mass rename** — `formatjs_library(name = \"dist\", …)` → `name = \"{basename}\"` and `formatjs_test(name = \"unit_test\", …)` → `name = \"{basename}_test\"` across ~38 BUILDs to match the plugin's defaults (the `ts_library_name` / `ts_test_name` directives are dropped).

- **Removed wasm targets** in `crates/icu_messageformat_parser/BUILD.bazel`: dropped `_wasm_lib` (`rust_shared_library`) and `formatjs_icu_messageformat_parser_wasm` (`rust_wasm_bindgen`), removed `lib_wasm.rs`, dropped `wasm-bindgen` / `serde-wasm-bindgen` from the crate's `Cargo.toml`, switched `crate-type` back to `[\"rlib\"]`. Nothing in this repo consumes the wasm artifact today.

- **Per-package opt-outs** — added `# gazelle:ts_enabled false` to ~9 directories whose layouts don't fit the standard library + test pattern (eslint-plugin-formatjs's hand-tuned external_packages, editor's `.tsx` entry, tsd type tests, integration-tests with bespoke `vitest()` rules, cli/integration-tests with intentionally-invalid TS fixtures, etc.).

- **Directives** — replaced 3 `# gazelle:formatjs_enabled false` with `# gazelle:ts_enabled false`.

## Test plan

- [x] \`bazel test //packages/... //docs/... //crates/...\` — 383 / 383 pass
- [x] \`bazel run //:gazelle\` is idempotent (no diff on a second run)
- [x] Manual \`ts_project\` rules in \`docs/\`, \`packages/eslint-plugin-formatjs/\`, \`packages/react-intl/examples/\` build correctly via the typecheck-shim path
- [x] One-off \`js_test\` in \`packages/cli-lib/integration-tests/\` still runs as a stock js_test
- [ ] CI: \`bazel test //...\` on Ubuntu + macOS


BUILD.bazel
```diff
@@ -12,6 +12,30 @@ load("//tools:tsconfig.bzl", "BASE_TSCONFIG")
...
 # gazelle:js_validate_import_statements off
 
+# gazelle_ts_plugin (from @gazelle_plugins//gazelle_ts_plugin) emits stock
+# ts_project + js_test rules; map_kind rewrites them to our wrappers, which
+# accept the same generated attrs (srcs/deps for libraries, data/entry_point
+# for tests) and translate to ts_project / vitest internally.
+# gazelle:map_kind ts_project formatjs_library //tools:compile.bzl
+# gazelle:map_kind js_test formatjs_test //tools:test.bzl
+# gazelle:ts_library_name dist
+# gazelle:ts_test_name unit_test
```
longlho: not necessary

BUILD.bazel
```diff
@@ -12,6 +12,30 @@ load("//tools:tsconfig.bzl", "BASE_TSCONFIG")
 # gazelle:js_pnpm_lockfile pnpm-lock.yaml
 # gazelle:js_validate_import_statements off
 
+# gazelle_ts_plugin (from @gazelle_plugins//gazelle_ts_plugin) emits stock
+# ts_project + js_test rules; map_kind rewrites them to our wrappers, which
+# accept the same generated attrs (srcs/deps for libraries, data/entry_point
+# for tests) and translate to ts_project / vitest internally.
+# gazelle:map_kind ts_project formatjs_library //tools:compile.bzl
+# gazelle:map_kind js_test formatjs_test //tools:test.bzl
+# gazelle:ts_library_name dist
```
longlho: shouldn't be necessary

BUILD.bazel
```diff
@@ -12,6 +12,30 @@ load("//tools:tsconfig.bzl", "BASE_TSCONFIG")
...
+# gazelle:ts_test_name unit_test
+# gazelle:ts_npm_link_pattern //:node_modules/{pkg}
+# gazelle:ts_project_references false
```
longlho: not necessary

BUILD.bazel
```diff
@@ -12,6 +12,30 @@ load("//tools:tsconfig.bzl", "BASE_TSCONFIG")
...
+# gazelle:ts_test_entry_point_auto false
```
longlho: not necessary

longlho:
Addressed your review:

**Directives dropped** — `ts_library_name`, `ts_test_name`, `ts_project_references=false`, `ts_test_entry_point_auto=false` all gone from root BUILD. Plugin defaults now win:
- libraries become `:basename` (e.g. `:intl-localematcher`), tests become `:basename_test`.
- `formatjs_library` silently absorbs `composite/declaration/declaration_map/source_map` since the rolldown / oxc pipelines don't use them.

**Why some packages still need `# gazelle:ts_enabled false`:**
- `eslint-plugin-formatjs` — hand-tuned `external_packages` list and a manual `copy_to_bin(name="srcs")` that conflicts with the macro's auto-generated `:srcs`.
- `editor` — entry is `index.tsx`, not the `index.ts` default.
- `cli/integration-tests` — `extract/*.tsx` fixtures are intentionally invalid TS used as CLI test input.
- `icu-messageformat-parser/integration-tests` — `run-parser.ts` is the entry of an existing `js_run_binary`; auto-generation would race on srcs.
- `react-intl/tests/tsd`, `intl/tests/tsd` — `*.test-d.ts` go through `tsd_test`, not vitest.
- `babel-plugin-formatjs/integration-tests`, `unplugin/integration-tests`, `ts-transformer/integration-tests` — bespoke `vitest()` rules with custom glob/fixtures; auto-generated `formatjs_test` would race.
- `cli-lib/integration-tests` — one-off ESM smoke test (just `entry_point` + a single npm dep). The macro now has a pass-through that forwards back to stock `js_test` for this shape, so it keeps working.
- `docs` — manual `copy_to_bin(name="srcs")` over both `src/**` and `public/**` assets.

These are all "the package's BUILD doesn't fit the standard library + test pattern" cases. The integration-tests opt-outs could be reclaimed by replacing the manual `vitest()` calls with `formatjs_test`, but that's a separate cleanup PR.

**Test wrapper fix:** the bundled `:basename` js_library on published packages includes the package's own `package.json`, which shadows the root `//:package.json` and breaks Node's `#packages/*` subpath resolution at runtime. The `formatjs_test` wrapper now strips the plugin-emitted sibling-library label and substitutes `:lib` (source-only) instead.

`bazel test //...` — 383 / 383 pass.

PR #6460, opened 2026-04-28
fix(formatjs_cli): extract formatMessage from callbacks inside optional chains

### LINKED ISSUE #6459: `formatjs_cli extract` misses `formatMessage()` calls nested in callbacks inside optional chains
## Which package?

`//crates/formatjs_cli`

## Describe the bug

The formatjs CLI does not identify messages like this: 

```ts
  // Case 6: formatMessage nested in a callback inside an optional chain.
  getUsers()?.map(() =>
    intl.formatMessage(
      {
        defaultMessage: 'In a callback inside an optional chain',
        description: 'Test callbacks inside an optional chain',
      },
    ),
  ).join(', ')
```

### Codesandbox URL

N/A

### Reproducible Steps/Repo

1. Go to `packages/ts-transformer/tests/fixtures/optionalChaining.tsx`
2. Add the failing test case to the fixture
3. Go to `crates/formatjs_cli/src/extractor.rs`
4. Add the expected message to the test: 
    ```rust
            MessageDescriptor {
                id: None,
                default_message: Some("In a callback inside an optional chain".to_string()),
                description: Some(Value::String("Test callbacks inside an optional chain".to_string())),
                file: None,
                start: None,
                end: None,
            },
    ```
5. Run `bazel test //crates/formatjs_cli:formatjs_cli_test`
6. Observe that the test fails

## Expected behavior

Expected behavior is that the Rust CLI extracts the message.

## Screenshots

N/A

- OS: [e.g. iOS] MacOS 
- Browser [e.g. chrome, safari] Chrome
- Version [e.g. 22] `master` (`c281a36db86d4f8e3370010bc568dc9852b59e5e`) 

## Additional context

Initially encountered while `using rules_formatjs@0.5.0`

walkerburgin:
Fixes #6459.

The Rust CLI's chain-expression visitor walked into call expression arguments, but skipped the callee. As a result, `formatMessage` calls inside callbacks passed to optional-chain methods (e.g. `getUsers()?.map(() => intl.formatMessage(...))`) were never extracted.

## Changes
- Walk both the callee and arguments of a `CallExpression` chain element so nested `formatMessage` calls inside callback bodies are extracted.
- Drop the `StaticMemberExpression` arm — `walk_chain_expression` covers it via the default case.
- Add fixture case in `packages/ts-transformer/tests/fixtures/optionalChaining.tsx`.
- Add Rust unit-test expectation in `crates/formatjs_cli/src/extractor.rs`.
- Add ts-transformer assertion in `packages/ts-transformer/tests/index.test.ts`.
- Add CLI conformance case `22_callback_in_optional_chain.txt` so both Rust and TS extractors are guarded.

## Test plan
- [x] `bazel test //crates/formatjs_cli:formatjs_cli_test`
- [x] `bazel test //packages/ts-transformer:ts-transformer_test`
- [x] `bazel test //packages/cli/integration-tests:conformance_test`

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -113,7 +113,8 @@
     "https://bcr.bazel.build/modules/jsoncpp/1.9.6/source.json": "a04756d367a2126c3541682864ecec52f92cdee80a35735a3cb249ce015ca000",
     "https://bcr.bazel.build/modules/libpfm/4.11.0/MODULE.bazel": "45061ff025b301940f1e30d2c16bea596c25b176c8b6b3087e92615adbd52902",
     "https://bcr.bazel.build/modules/llvm/0.7.3/MODULE.bazel": "b8ffe1ec9d18c0deb6d6dc1a6f78a64bcfcec8648c97b92ed6cd98a7fe49c64f",
-    "https://bcr.bazel.build/modules/llvm/0.7.3/source.json": "44d694e1559683ea4cf96052c56c5426d2604016636133798fcd57cc78542984",
+    "https://bcr.bazel.build/modules/llvm/0.7.5/MODULE.bazel": "5a2b497560edcd2d747f9b1d4df18f79c92834cf885e9a54e129654b03de7f99",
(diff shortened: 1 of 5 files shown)


walkerburgin:
Thank you!

PR #6466, opened 2026-04-29
fix(@formatjs/intl-durationformat): use BigDecimal for sub-second rollups

### LINKED ISSUE #6462: DurationFormat introduces error in fractional digits
## Which package?

@formatjs/intl-durationformat

## Describe the bug

DurationFormat introduces what appears to be a floating precision error when given specific durations. What it appears to be doing under the hood is the equivalent of `second + millisecond`, and indeed any such numbers that produce the same floating precision error in JS such as `1 + .118` show up with the same error. However, since duration units are given in whole numbers there should be no reason to introduce floating arithmetic, and consequently floating error, into formatting.

## To Reproduce

https://jsfiddle.net/chsk6t42/

### Reproducible Steps/Repo

```js
new DurationFormat('en', {
  milliseconds: 'numeric',
  style: 'narrow',
}).format({ milliseconds: 473, seconds: 1 });
```
Outputs the following: `1.472999999s`

## Expected behavior

Intl.DurationFormat on Chrome 147 outputs: `1.473s`

- OS: Windows 11
- Browser: Chrome
- Version: 147

longlho:
## Summary
Sub-second units that roll up into a numeric parent (`milliseconds: 'numeric'`, etc.) were combined with float arithmetic. `1 + 473/1e3` lands on `1.4729999999999998650`, and `roundingMode: 'trunc'` truncated that to `1.472999999s` instead of `1.473s`.

Carry the value as `BigDecimal` end-to-end and pass its decimal string to `NumberFormat`. NumberFormat (V3, ES2023) parses the string as a Mathematical Value, sidestepping the IEEE 754 round-trip entirely.

Also adds `ES2023.Intl` to `BASE_TSCONFIG.lib` so `formatToParts` accepts the `${number}` string overload at type-check time.

Fixes #6462

## Test plan
- [x] New regression test covers the seconds/milliseconds/microseconds rollup cases
- [x] `bazel test //packages/intl-durationformat/...` — all 7 tests pass
- [x] `bazel test //packages/...` — full suite (374 tests) passes; the `ES2023.Intl` lib addition didn't break any other package's typecheck
- [ ] CI green

### DIFF
--- a/packages/intl-durationformat/abstract/PartitionDurationFormatPattern.ts
+++ b/packages/intl-durationformat/abstract/PartitionDurationFormatPattern.ts
@@ -34,9 +34,7 @@ export function PartitionDurationFormatPattern(
     // Carry the value as BigDecimal end-to-end so sub-second rollups stay
     // exact. Float arithmetic like `1 + 473/1e3` lands on
     // `1.4729999999999998650`, which `roundingMode: 'trunc'` truncates to
-    // `1.472999999` instead of `1.473` (#6462). NumberFormat (V3) accepts
-    // a decimal string and parses it as a Mathematical Value, sidestepping
-    // the IEEE 754 round-trip entirely.
+    // `1.472999999` instead of `1.473` (#6462).
     let value = new BigDecimal(duration[row.valueField])
     const style = internalSlots[row.styleSlot]
     const display = internalSlots[row.displaySlot]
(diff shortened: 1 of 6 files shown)


packages/intl-durationformat/abstract/PartitionDurationFormatPattern.ts
```diff
@@ -95,7 +104,7 @@ export function PartitionDurationFormatPattern(
           value: separator,
         })
       }
-      let parts = nf.formatToParts(value)
+      let parts = nf.formatToParts(value.toString() as `${number}`)
```
longlho: this is lossy, should carry BigDecimal all the way through

longlho: might need to type guard downstream

longlho:
Addressed in 17c55a96c — now passes `BigDecimal` straight through to `nf.formatToParts` (no `.toString()`, no `${number}` cast). Native NumberFormat coerces non-primitive inputs via ToPrimitive→toString into a StringNumericLiteral, so the exact-decimal semantics survive without an intermediate string round-trip. Widened `createMemoizedNumberFormat`'s return type (`NumberFormatLike`) to declare `format`/`formatToParts` as accepting `BigDecimal | string` so the call typechecks. Reverted the `ES2023.Intl` lib bump since it was only needed by the old cast.

PR #6545, opened 2026-05-12
fix(@formatjs/intl-numberformat): expose rounding resolved options

longlho:
## Summary
- include roundingIncrement, roundingMode, and trailingZeroDisplay in Intl.NumberFormat.prototype.resolvedOptions()
- add coverage for explicit and default rounding resolved options

## Spec
- ECMA-402: [Intl.NumberFormat.prototype.resolvedOptions](https://tc39.es/ecma402/#sec-intl.numberformat.prototype.resolvedoptions)
- ECMA-402: [Resolved Options of NumberFormat Instances](https://tc39.es/ecma402/#table-intl-numberformat-resolvedoptions)

## Test
- bazel test //packages/intl-numberformat:misc_tests

### DIFF
--- a/packages/intl-numberformat/tests/misc.test.ts
+++ b/packages/intl-numberformat/tests/misc.test.ts
@@ -334,9 +334,12 @@ it('correctly set default options', () => {
     minimumIntegerDigits: 1,
     notation: 'standard',
     numberingSystem: 'latn',
+    roundingIncrement: 1,
+    roundingMode: 'halfExpand',
     roundingPriority: 'auto',
     signDisplay: 'auto',
     style: 'decimal',
+    trailingZeroDisplay: 'auto',
     useGrouping: 'auto',
   })
 })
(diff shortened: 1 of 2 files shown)


packages/intl-numberformat/tests/misc.test.ts
```diff
@@ -438,6 +441,21 @@ test('#4359 roundingIncrement with fraction digits', () => {
...
+test('resolvedOptions includes rounding options', () => {
+  const resolved = new NumberFormat('en-US', {
+    roundingIncrement: 5,
+    roundingMode: 'ceil',
+    trailingZeroDisplay: 'stripIfInteger',
+    maximumFractionDigits: 2,
+    minimumFractionDigits: 2,
+  }).resolvedOptions()
+
+  expect(resolved.roundingIncrement).toBe(5)
```
longlho, line 454: need actual format output test

PR #6549, opened 2026-05-12
fix(@formatjs/intl-durationformat): align fractional unit style

longlho:
## Summary
- add the current GetDurationUnitOptions twoDigitHours parameter shape
- use an internal fractional style for sub-second units rolled into a preceding numeric unit
- keep resolvedOptions reporting public numeric values for fractional sub-second slots

## Spec
- ECMA-402: [GetDurationUnitOptions](https://tc39.es/ecma402/#sec-getdurationunitoptions)
- ECMA-402: [PartitionDurationFormatPattern](https://tc39.es/ecma402/#sec-partitiondurationformatpattern)

## Test
- bazel test //packages/intl-durationformat:intl-durationformat_test

### DIFF
--- a/packages/intl-durationformat/tests/index.test.ts
+++ b/packages/intl-durationformat/tests/index.test.ts
@@ -84,6 +84,12 @@ test('Intl.DurationFormat format digital', function () {
     new DurationFormat('en', customFormatterOptions).resolvedOptions()
       .milliseconds
   ).toBe('numeric')
+  expect(
+    new DurationFormat('en', customFormatterOptions).format({
+      seconds: 1,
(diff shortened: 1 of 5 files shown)


packages/intl-durationformat/tests/index.test.ts
```diff
@@ -80,6 +80,10 @@ test('Intl.DurationFormat format digital', function () {
   expect(
     new DurationFormat('en', customFormatterOptions).format({years: 1})
   ).toBe('1 yr, 0:00:00')
+  expect(
```
longlho, line 83: need test for actual output of fractional

PR #6834, opened 2026-06-27
chore(deps): update typescript to v7 rc

longlho:
## Summary

- bump root TypeScript to `7.0.1-rc`
- update the `rules_ts` TypeScript integrity and module lock
- wire the TypeScript 7 RC optional platform packages into `pnpm-lock.yaml`
- allow the affected package BUILD metadata to accept TypeScript 7

## Validation

- `bazel mod deps`
- `git diff --check`

Known failures:

- `bazel test //packages/ts-transformer:typecheck_typecheck_test`
  - `aspect_rules_ts` creates an isolated `@npm_typescript` repo that only contains the `typescript` package. TypeScript 7 RC's CLI loads `@typescript/typescript-darwin-arm64`, so `tsc` cannot start from that repo.
  - TypeScript 7 RC no longer exports the package root, so validator/app imports that resolve `typescript` fail with `ERR_PACKAGE_PATH_NOT_EXPORTED`.
- `pnpm install --lockfile-only --frozen-lockfile`
  - the refreshed main lockfile currently has a duplicate mapping key, so pnpm rejects it before it can validate this bump.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -78,7 +78,7 @@ bazel_dep(name = "aspect_rules_ts", version = "3.8.9")
 
 rules_ts_ext = use_extension("@aspect_rules_ts//ts:extensions.bzl", "ext", dev_dependency = True)
 rules_ts_ext.deps(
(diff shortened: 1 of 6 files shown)


longlho:
Closing this direct replacement. TypeScript 7 no longer exposes the JavaScript compiler API that FormatJS and rules_ts still require, and the isolated Bazel validator cannot see the platform binary package. The safe migration is a separate side-by-side change: keep TypeScript 6 for API consumers and wire TypeScript 7 only as the tsc CLI.

PR #6855, opened 2026-07-10
build(@formatjs/intl-datetimeformat): make tzdata generation hermetic

longlho:
## Summary
- build IANA zic and zdump from source with target-scoped hermetic LLVM on macOS and Linux
- keep zic and zdump in the execution configuration
- replace shell and coreutils generation with pinned Node actions
- validate timezone generation directly on Ubuntu and macOS CI
- preserve explicit labels for all generated timezone outputs

## Why
Global LLVM registration can be overridden by caller-provided `--extra_toolchains`; zic then switches to the host compiler. The old generation path also depended on host Bash and coreutils and kept an unused `/usr/sbin/zic` script.

## Validation
- `pnpm t`
- `bazel build //packages/intl-datetimeformat:pkg`
- macOS timezone builds with both normal and competing GCC toolchain selection
- all 427 generated zdump files are byte-identical to the previous pipeline

Follow-up to #6854.


.github/workflows/test.yml
```diff
@@ -190,6 +190,17 @@ jobs:
         with:
           node-version: '24'
 
+      - name: Hermetic Timezone Cross-Build
+        run: |
+          $ErrorActionPreference = 'Stop'
+          $PSNativeCommandUseErrorActionPreference = $true
+          $bazelTimezoneFlags = @(
+            '--platforms=@llvm//platforms:linux_aarch64_gnu.2.28',
```
longlho: this shouldn't have to be explicit, we can auto-detect host

packages/intl-datetimeformat/BUILD.bazel
```diff
@@ -161,7 +161,14 @@ generate_tz_data(
 
 filegroup(
     name = "zdumps",
-    srcs = ["tz_data_generated/zdump/%s" % zone for zone in ZONES],
```
longlho, line 164: this makes sure outputs are explicit, if there are new zones there should be a separate write_source_files target to golden test this

PR #6946, opened 2026-07-31
Documented how to use ember-intl

### LINKED ISSUE #6931: Document how to use ember-intl
## Background

In June 2024, I had opened the pull request #4458 so that people from the wider JavaScript community can see how to set up and use [`ember-intl`](https://github.com/ember-intl/ember-intl). I had closed the PR because there wasn't a response for a couple of months.

Since then, the Ember community has been moving to using Vite to build Ember apps (so-called "v2 apps"). Accordingly, we found a way to make `ember-intl` keep working on v2 apps.

## Proposal

I'd like to update `formatjs`'s documentation site, similarly to #4458 but with the latest information. Concretely, I would:

1. Update https://formatjs.github.io/docs/getting-started/installation/#minimal-application to show an example for Ember.
1. Create https://formatjs.github.io/docs/ember-intl with contents similar to what's shown in https://formatjs.github.io/docs/vue-intl. For brevity, I'd only address v2 apps and provide a link to [`ember-intl`'s documentation site](https://ember-intl.github.io/ember-intl/) for more information.

To limit the scope, the pull request for this issue won't address how to use `formatMessage` with `ember-intl`. The documentation site for `ember-intl` covers this case. That is, the following pages, which mentioned Vue, won't be modified to also mention Ember:

- https://formatjs.github.io/docs/getting-started/message-declaration
- https://formatjs.github.io/docs/getting-started/message-extraction
- https://formatjs.github.io/docs/getting-started/message-distribution

ijlee2:
## Background

Closes #6931. I updated `formatjs`'s documentation site, similarly to #4458 but with the latest information.

1. Updated https://formatjs.github.io/docs/getting-started/installation/#minimal-application to show an example of `formatMessage` and `formatNumber` for Ember.
1. Created https://formatjs.github.io/docs/ember-intl to show the minimal setup for v2 apps and examples of the `intl` service and `t` helper.

Note, I couldn't run and test the docs site locally (some problem with React hooks). I also wasn't sure if files like `docs/public/sitemap.xml` and `docs/src/docs-metadata.generated.json` are auto-generated or to be manually updated.

### DIFF
--- a/docs/src/docs/ember-intl.mdx
+++ b/docs/src/docs/ember-intl.mdx
@@ -81,6 +81,20 @@ export default class ApplicationRoute extends Route {
 }
 ```
 
+File paths prefixed with `virtual:` are called a "virtual module" in Vite. They don't physically exist on disk. If you use TypeScript: Add the path `@ember-intl/vite/virtual` to `compilerOptions.types` in `tsconfig.json` so that TypeScript can understand the virtual modules from `ember-intl`.
+
+```json
+/* tsconfig.json */
+{
+  "compilerOptions": {
+    "types": [
+      /* ... */
+      "@ember-intl/vite/virtual"
+    ]
+  }
+}
+```
+
 For more information, see [Quickstart](https://ember-intl.github.io/ember-intl/docs/quickstart) and [Lazy-loading translations](https://ember-intl.github.io/ember-intl/docs/advanced/lazy-loading-translations).
 
 ## Usage
--- a/docs/src/docs/ember-intl.mdx
+++ b/docs/src/docs/ember-intl.mdx
@@ -1,0 +1,177 @@
+`ember-intl` helps you internationalize Ember apps and addons.
+
+## Installation
+
+This page shows the minimal setup for "v2 apps" (Ember apps built with Vite).
+
+<Tabs
+groupId="npm"
+defaultValue="npm"
+values={[
+{label: 'npm', value: 'npm'},
+{label: 'yarn', value: 'yarn'},
+{label: 'pnpm', value: 'pnpm'},
+]}>
+<TabItem value="npm">
+
+```sh
+npm i -S ember-intl @ember-intl/vite
+```
+
+</TabItem>
+<TabItem value="yarn">
+
+```sh
+yarn add ember-intl @ember-intl/vite
+```
+
+</TabItem>
+<TabItem value="pnpm">
+
+```sh
+pnpm add ember-intl @ember-intl/vite
+```
+
+</TabItem>
+</Tabs>
+
+To define translations, create a JSON or YAML file in the `translations` folder.
+
+```json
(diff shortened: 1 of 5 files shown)


docs/src/docs/getting-started/installation.mdx
```diff
@@ -41,11 +41,86 @@ After following the step above, you should be able to get a minimal application
 groupId="engine"
 defaultValue="react"
 values={[
+{label: 'Ember', value: 'ember'},
 {label: 'Node', value: 'node'},
 {label: 'React', value: 'react'},
 {label: 'Vue3', value: 'vue'},
 ]}>
 
+<TabItem value="ember">
```
longlho, line 50: This tab won't run after the install step above, since that only installs `react` and `react-intl`. Can we add the `ember-intl` / `@ember-intl/vite` install step here, or link to the Ember guide?

docs/src/docs/ember-intl.mdx
```diff
@@ -0,0 +1,177 @@
...
+})
+```
+
+When the app starts, `@ember-intl/vite` automatically loads your translations, but it does not pass them to the `intl` service (the source of truth). You are responsible for importing the translations that you need, then calling `addTranslations` to pass them to the `intl` service.
+
+```ts
+/* app/routes/application.ts */
+import Route from '@ember/routing/route'
+import {type Registry as Services, service} from '@ember/service'
+import translationsForEnUs from 'virtual:ember-intl/translations/en-us'
```
longlho, line 68: `@ember-intl/vite` also needs `@ember-intl/vite/virtual` in `compilerOptions.types` for TypeScript. Can we include the `tsconfig.json` step so this example type-checks?

longlho (commented):
Two setup gaps to fix:

docs/src/docs/getting-started/installation.mdx
ijlee2: Hmm, wouldn't that be the case for the Node and Vue examples as well? I believe these would require `@formatjs/intl` and `vue-intl`, respectively.

Maybe we could instead take one of these two actions? I'd prefer the second as it is more actionable (easier to implement).

1. Add a sentence to the subsection "Installation," providing internal links to the Ember and Vite pages for more information.
1. Remove the sentence "After following the step above, you should be able to get a minimal application like this running:"

docs/src/docs/ember-intl.mdx
ijlee2: Thanks, done.

docs/src/docs/getting-started/installation.mdx
longlho: yeah that's fair I can fix them holistically later

ijlee2: Sounds good, appreciate your help.

PR #7005, opened 2026-08-07
fix(formatjs_cli): fail on input resolution errors

### LINKED ISSUE #7004: [cli-native] extract exits 0 with empty/partial catalog when input paths fail to resolve, even with --throws
## Which package?

Native CLI (`crates/formatjs_cli`) — reproduced on `formatjs_cli` 1.2.1 release binary; the code path exists on current `main`. The TypeScript CLI (`@formatjs/cli` 6.16.16) has the same behavior for missing paths, but this report focuses on the native CLI because it's the one wired into build systems via `rules_formatjs`, where the consequences are severe.

## Describe the bug

`extract` silently succeeds (exit 0) with an **empty or partially truncated catalog** when input paths fail to resolve — even with `--throws`. There is no error, no warning, nothing on stderr.

Three silent-drop paths in `resolve_files_from_globs` ([extract.rs](https://github.com/formatjs/formatjs/blob/6e356ac9d9efd5162dd808131291f87fb61f6092/crates/formatjs_cli/src/extract.rs#L247-L267)):

1. An explicit input file whose `stat` fails (missing, unreadable parent dir, or a *transient* I/O error such as `EMFILE`/`ENOMEM` under load) returns `false` from `glob_path.is_file()` — the path then falls through to glob handling instead of erroring (L247).
2. If the derived base dir also fails `exists()`, the path is skipped with a bare `continue` (L260).
3. All `WalkDir` traversal errors are discarded via `.filter_map(|e| e.ok())` (L267).

`--throws` only covers per-file parse/extract errors *after* file resolution, so none of these are catchable.

## Reproduction

```sh
# formatjs_cli-linux-x64 1.2.1 release binary

cat > src/App.tsx <<'EOF'
import {defineMessages} from 'react-intl';
export const messages = defineMessages({
  greeting: {id: 'app.greeting', defaultMessage: 'Hello, world'},
});
EOF
mkdir locked && cat > locked/Locked.tsx <<'EOF'
import {defineMessages} from 'react-intl';
export const locked = defineMessages({
  secret: {id: 'locked.secret', defaultMessage: 'Restricted area'},
});
EOF

# 1. Missing explicit file → exit 0, empty catalog
$ formatjs extract src/DoesNotExist.tsx --throws --out-file out.json; echo $?
0
$ cat out.json
{}

# 2. Existing file behind an unreadable dir → exit 0, empty catalog
$ chmod 000 locked
$ formatjs extract locked/Locked.tsx --throws --out-file out.json; echo $?
0
$ cat out.json
{}

# 3. Mixed list → exit 0, *silently truncated* catalog
$ formatjs extract src/App.tsx locked/Locked.tsx --throws --out-file out.json; echo $?
0
$ cat out.json
{
  "app.greeting": {
    "defaultMessage": "Hello, world"
  }
}
```

The same command with `chmod 755 locked` produces the complete 2-message catalog with the same exit code — i.e. identical declared inputs yield both truncated and complete "successful" outputs depending on transient filesystem state.

## Real-world impact: build-cache poisoning

We run the native CLI (1.1.20) through `rules_formatjs` 0.9.0 in a large Bazel monorepo with a shared remote cache. `formatjs_extract` passes an explicit list of source files on the command line. When a transient filesystem hiccup on a CI host makes `is_file()` fail for some inputs, the action still exits 0 and Bazel caches the truncated catalog as a valid result for that action key — permanently, for every subsequent build with cache read access.

We traced a translation-pipeline incident to exactly this: cached `FormatjsExtract` outputs that were empty (or retained 18 of 2,289 message IDs) while forced local re-execution of the same action produced complete catalogs. Downstream, a catalog diff between two adjacent commits — whose sole change was CSS classNames — reported **73,916 phantom translation changes** (2,174 "new" strings × 34 locales).

Since #6611 (1.1.8+), extraction fan-out via Rayon defaults to all visible logical CPUs, which increases pressure on CI hosts running many concurrent extraction actions and makes these transient resolution failures more likely; but the dropping mechanism itself is the silent error swallowing above, which predates the parallelization.

## Expected behavior

- With `--throws` (arguably always), an explicit non-wildcard input path that fails to resolve should be a hard error, like the TS compiler treats a missing input. Distinguishing `Err` from `Ok(false)` on `fs::metadata` would also separate "doesn't exist" from "couldn't stat".
- `WalkDir` errors should at minimum be reported to stderr, and fail the run under `--throws`.
- Consider logging the resolved-file count so wrappers can sanity-check `inputs declared == files processed`.

Happy to send a PR if the maintainers agree on the direction.

## Environment

- `formatjs_cli` 1.2.1 (linux-x64 release binary), also observed with 1.1.20 via `rules_formatjs` 0.9.0
- Linux x86_64

longlho:
## Summary

- distinguish literal inputs from glob patterns before traversal
- validate positional and `--in-file` inputs through the same resolver
- fail on missing inputs, unsupported literal files, inaccessible glob bases, and traversal errors when `--throws` is enabled
- keep best-effort behavior without `--throws`, but report filesystem errors to stderr
- document the stricter native CLI behavior

Fixes #7004.

## Compatibility

`extract "optional/**/*.ts" --throws` now fails when the glob base directory does not exist. Callers relying on absent optional directories should create the directory or omit `--throws`.

## Validation

- `bazel test //crates/formatjs_cli:formatjs_cli_test`
- direct `formatjs extract missing.tsx --throws` exits 1 with the filesystem error
- direct extraction without `--throws` exits 0 and reports the filesystem warning

### DIFF
--- a/crates/formatjs_cli/README.md
+++ b/crates/formatjs_cli/README.md
@@ -244,7 +244,7 @@ formatjs extract "src/**/*.{js,ts,tsx}" \
 - `--additional-component-names <NAMES>` - Additional component names to extract from (comma-separated)
 - `--additional-function-names <NAMES>` - Additional function names to extract from (comma-separated)
 - `--ignore <PATTERNS>` - Glob patterns to exclude
-- `--throws` - Throw exception when failing to process any file
+- `--throws` - Exit with an error when input resolution, traversal, or extraction fails
 - `--pragma <PRAGMA>` - Parse custom pragma for file metadata (e.g., `@intl-meta`)
 - `--preserve-whitespace` - Preserve whitespace and newlines
 - `--flatten` - Hoist selectors and flatten sentences
--- a/crates/formatjs_cli/src/extract.rs
+++ b/crates/formatjs_cli/src/extract.rs
@@ -86,7 +86,7 @@ pub fn extract_to_string(
     let file_list = if let Some(in_f) = in_file {
         read_file_list(in_f)?
     } else {
-        resolve_files_from_globs(files, ignore, follow_links)?
+        resolve_files_from_globs(files, ignore, follow_links, throws)?
     };
 
     // Step 2: Extract messages from all files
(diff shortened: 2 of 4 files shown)


longlho:
This closes all three silent-drop paths from #7004 — literal inputs stat'd via `fs::metadata` with `Err` handled, base dir via `try_exists()`, and `WalkDir` errors no longer swallowed. The test asserting the output file isn't written on failure is exactly the contract build systems need. Two behavior changes worth calling out in release notes, plus two smaller notes (inline review is blocked for this account, so line refs below):

**[L286](https://github.com/formatjs/formatjs/blob/19d7b7ca082792868ee43d0cc5d3cd14089329bf/crates/formatjs_cli/src/extract.rs#L286) — nonexistent glob base under `--throws`:** this is a behavior change worth flagging in the changelog — `extract "optional/**/*.ts" --throws` previously succeeded with an empty result and now hard-fails. Correct for build-system use, but CI setups that glob over optional directories will break on upgrade. Nit for symmetry: the `Err` arm reports to stderr in non-throws mode but this arm continues silently — if intentional (globs matching nothing is normal), a short code comment would help.

**[L254](https://github.com/formatjs/formatjs/blob/19d7b7ca082792868ee43d0cc5d3cd14089329bf/crates/formatjs_cli/src/extract.rs#L254) — filtered literal inputs:** a literal input that exists but fails `is_supported_file` is still dropped silently, even with `--throws`. For explicit argv file lists that's likely a user/wrapper error rather than filtering intent — worth at least an `eprintln!` (staying silent for `should_ignore` matches makes sense).

**[L251](https://github.com/formatjs/formatjs/blob/19d7b7ca082792868ee43d0cc5d3cd14089329bf/crates/formatjs_cli/src/extract.rs#L251) — glob metachars in literal names:** nice ordering — because `fs::metadata` runs before the `is_glob` branch, an existing file whose name contains glob metachars (e.g. `foo[1].ts`) still resolves as a literal file. Worth a small test to lock that in so a future refactor doesn't reorder the checks.

**[L197](https://github.com/formatjs/formatjs/blob/19d7b7ca082792868ee43d0cc5d3cd14089329bf/crates/formatjs_cli/src/extract.rs#L197) — `--in-file` parity:** entries from `read_file_list` don't get the same strictness — a missing file in the list still surfaces later at read time with a less specific error. Fine to punt, but parity here would be a natural follow-up since "failed to read" is less actionable than "failed to resolve input file".

longlho:
Addressed in `5cc4bb8a4`:

- unsupported literal inputs now error with `--throws` and warn otherwise
- `--in-file` entries now use the same resolution path as positional inputs
- best-effort handling for a nonexistent glob base now has an explicit code comment
- PR description now calls out the optional-directory compatibility change

The glob-metachar ordering already has coverage in `test_resolve_files_literal_path_containing_glob_metacharacters`, so I kept that test in place.

Validation: `bazel test //crates/formatjs_cli:formatjs_cli_test`.

PR #7182, opened 2026-09-05
fix(formatjs_cli): extract messages from tagged template literals

vpontis:
`@formatjs/cli-lib` 9.0.5 switched to native extraction (#7046). The Rust extractor rejects a tagged no-substitution template such as

```ts
defaultMessage: dedent`
  Hi {name},
  your ticket is ready.
`,
```

with `` `defaultMessage` must be a string literal to be extracted ``, so every message written that way silently dropped out of the catalog. `@formatjs/ts-transformer` and `eslint-plugin-formatjs` both accept these (the `templateLiteral.tsx` fixture and its snapshot already cover `dedent`), so this restores parity.

The first commit tightens `test_fixture_template_literal`, which only asserted that at least one message was extracted, and adds an inline test for tagged templates in function calls, JSX attributes and descriptions. Both fail on `main`. The second commit adds the `TaggedTemplateExpression` arm. A tagged template with substitutions still falls through to the existing "must be a string literal" error.

Tested with `cargo test -p formatjs_cli` (239 passed). I don't have Bazel set up locally.

### DIFF
--- a/crates/formatjs_cli/src/extractor.rs
+++ b/crates/formatjs_cli/src/extractor.rs
@@ -290,8 +290,7 @@ impl<'a> MessageExtractor<'a> {
             Expression::TemplateLiteral(tpl) => {
                 self.extract_template_literal(tpl, preserve_whitespace)
             }
-            // Tagged templates like dedent`...` are treated as their raw template,
-            // matching @formatjs/ts-transformer and eslint-plugin-formatjs.
+            // The tag is ignored: dedent`...` extracts as its raw template, like ts-transformer.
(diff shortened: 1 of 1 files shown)


longlho (approved):

vpontis:
@longlho btw I was surprised to see the change to the Rust compiler was just a patch bump. I would have expected minor or major?

longlho:
It was a patch because it was fixing a bug in picking which extractor to use. It was supposed to swap to the rust one a while ago.
Thanks for your contribution.

PR #7183, opened 2026-09-06
feat(formatjs_cli): skip a message with a formatjs-ignore comment

vpontis:
We store notifications in a database and translate them on read, since one notification can be read in different locales. The message is defined once as a literal, which the extractor picks up:

```ts
notifT({ id: "qY2CVI", defaultMessage: "{inviter_name} added you to the check-in staff for {event_name}" }, values)
```

`notifT` saves the `id` and `defaultMessage` into the row. When a user fetches notifications, we format the stored row in their locale:

```ts
t({ id: notification.title_id, defaultMessage: notification.title_template }, values)
```

This is a render site with a dynamic id by design, but the extractor warns on every run:

```
notifications.ts:129:21 [FormatJS] `defaultMessage` must be a string literal to be extracted.
```

The only way to quiet it today is `onMsgError`, which is global and can't tell an intentional site from a mistake.

This adds a `formatjs-ignore` comment that skips the next call or JSX element, like `prettier-ignore`. Anything after the marker is a reason:

```tsx
// formatjs-ignore extracted where notifT is called
t({ id: notification.title_id, defaultMessage: notification.title_template }, values)

{/* formatjs-ignore */}
<FormattedMessage id={notification.title_id} defaultMessage={notification.title_template} />
```

The comment applies to a call or element starting on the next line or later on the same line. The message is skipped entirely: no extraction, no error. The marker must be followed by whitespace or end of comment, so `formatjs-ignore-file` could be added later.

Extractor only. `eslint-plugin-formatjs` rules still fire on these sites and could honor the same comment in a follow-up.

### DIFF
--- a/crates/formatjs_cli/src/extractor.rs
+++ b/crates/formatjs_cli/src/extractor.rs
@@ -7,7 +7,7 @@ use oxc_allocator::Allocator;
 use oxc_ast::ast::*;
 use oxc_data_structures::rope::{Rope, get_line_column};
 use oxc_parser::{Parser, ParserReturn};
-use oxc_span::SourceType;
+use oxc_span::{SourceType, Span};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


longlho:
This is not really a pattern we recommend. Whichever layer owns the template should format it, because there's no backwards compatibility for dynamic strings/ids.
Storing templates in db and letting clients assume certain placeholders exist is inherently unsafe.

PR #7186, opened 2026-09-06
codex/ecma402 coercion

longlho:
Fix Number-only input coercion from audit #7185, item 03. DateTimeFormat, RelativeTimeFormat, duration fields, and numeric options now reject BigInt, including boxed values and objects that return BigInt. Boolean duration fields convert normally.

Keep formatted plural operands on the exact Decimal path, preserving NumberFormat/PluralRules BigInt precision. Comments cite the relevant ECMA-262/402 algorithms.

Validation: six focused Bazel suites passed (shared abstracts, DateTimeFormat, RelativeTimeFormat, DurationFormat, NumberFormat, PluralRules), including coercion and unsafe-integer operand regressions.

### DIFF
--- a/packages/ecma262-abstract/ToNumber.ts
+++ b/packages/ecma262-abstract/ToNumber.ts
@@ -1,9 +1,15 @@
 import {Decimal} from '@formatjs/bigdecimal'
 
-/** https://tc39.es/ecma262/#sec-tonumber */
+/**
+ * ECMA-262 §7.1.4 ToNumber, steps 2, 8–10.
+ * https://tc39.es/ecma262/#sec-tonumber
+ * https://github.com/tc39/ecma262/blob/dcf59856a8184792a9e42f0ffb7dc064094a5dcc/spec.html#L5183-L5191
+ */
 export function ToNumber(arg: any): Decimal {
-  // Unary + evaluation step 3 returns ? ToNumber, preserving abrupt completions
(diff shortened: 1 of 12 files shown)


packages/ecma262-abstract/ToNumber.ts
```diff
@@ -1,47 +1,8 @@
...
-      return new Decimal(NaN)
-    }
-  }
-  invariant(typeof arg === 'object', 'object expected', TypeError)
-  let primValue = ToPrimitive(arg, 'number')
-  invariant(typeof primValue !== 'object', 'object expected', TypeError)
-  return ToNumber(primValue)
+  // Unary plus performs ToNumber, including object coercion and rejection of
+  // BigInt/Symbol. Decimal preserves that Number result for downstream arithmetic.
+  return new Decimal(+arg)
```
longlho, line 14: are we losing spec errors here?

longlho: Errors propagate unchanged before Decimal construction. [ECMA-262 §13.5.4.1, step 2](https://tc39.es/ecma262/#sec-unary-plus-operator-runtime-semantics-evaluation) applies `? ToNumber` ([source line 20666](https://github.com/tc39/ecma262/blob/dcf59856a8184792a9e42f0ffb7dc064094a5dcc/spec.html#L20666)). Added regressions for exact exception identity from coercion getters/methods, invalid primitive conversion, and BigInt/Symbol TypeErrors.

PR #7190, opened 2026-09-06
fix(deps): negotiate well-formed numbering system options

longlho:
Fix audit #7185, item 04. NumberFormat and DurationFormat now validate numberingSystem syntax before locale negotiation. Well-formed unsupported identifiers fall back; malformed identifiers still throw RangeError.

Extracts `IsUnicodeLocaleIdentifierType` into the shared abstract layer, also replacing duplicate DateTimeFormat and RelativeTimeFormat checks. Collator, DisplayNames, and Locale reuse it in their later PRs. Comments cite ECMA-402 §9.2.8 ResolveOptions, step 6.d.ii, with pinned spec-source lines.

Adds public/internal documentation and regressions for unsupported, case-insensitive, and malformed identifiers, including trailing line terminators.

Validation: focused Bazel suites passed. The full rebased stack passes all 139 polyfill/abstract checks and all 43 audit probes.


packages/ecma402-abstract/NumberFormat/InitializeNumberFormat.ts
```diff
@@ -58,10 +56,10 @@ export function InitializeNumberFormat(
   )
   if (
     numberingSystem !== undefined &&
-    numberingSystemNames.indexOf(numberingSystem) < 0
+    !/^[a-z0-9]{3,8}(-[a-z0-9]{3,8})*$/i.test(numberingSystem)
```
longlho: don't we have this validation in another abstract method in another polyfill already?

packages/ecma402-abstract/NumberFormat/InitializeNumberFormat.ts
```diff
@@ -58,10 +56,10 @@ export function InitializeNumberFormat(
...
   ) {
-    // 8.a. If numberingSystem does not match the Unicode Locale Identifier type nonterminal,
-    // throw a RangeError exception.
+    // Validate syntax; unsupported types fall back during ResolveLocale.
```
longlho: where in the spec is this specified? link is not specific enuf?

packages/ecma402-abstract/NumberFormat/InitializeNumberFormat.ts
longlho: Yes. Extracted `IsUnicodeLocaleIdentifierType` into `ecma402-abstract` and reused it in NumberFormat, DurationFormat, DateTimeFormat, and RelativeTimeFormat here. Collator, DisplayNames, and Locale reuse it in their later PRs. This removes the duplicate grammar checks without making other polyfills depend on Intl.Locale. Added valid/malformed type regressions, including trailing line terminators.

packages/ecma402-abstract/NumberFormat/InitializeNumberFormat.ts
longlho: [ECMA-402 §9.2.8 ResolveOptions, step 6.d.ii](https://tc39.es/ecma402/#sec-resolveoptions) requires the Unicode `type` grammar and throws RangeError for malformed values ([source line 314](https://github.com/tc39/ecma402/blob/b1c961988b9a07894b1dc3dc2b5626ea48387d61/spec/negotiation.html#L314)). Support is negotiated separately by §9.2.7 ResolveLocale, step 13.k. Comments now include the section, exact step, and pinned source-line link.

PR #7191, opened 2026-09-06
fix(@formatjs/intl-collator): validate collation and comparison inputs

longlho:
Validate `collation` against Unicode type syntax before locale negotiation. Use ECMA-262 ToString for comparison operands so Symbols throw and object coercion uses the string hint.

Closes audit findings 05–06 from #7185. Code comments link the relevant ECMA-402 clauses; unit tests cover malformed/unknown collations and comparison coercion.

Validation: `bazel test //packages/intl-collator:intl-collator_test`.

### DIFF
--- a/packages/intl-collator/core.ts
+++ b/packages/intl-collator/core.ts
@@ -1,3 +1,4 @@
+import {ToString} from '#packages/ecma262-abstract/ToString.js'
 import {OrdinaryHasInstance} from '#packages/ecma262-abstract/OrdinaryHasInstance.js'
 import {ToObject} from '#packages/ecma262-abstract/ToObject.js'
 import {CanonicalizeLocaleList} from '#packages/ecma402-abstract/CanonicalizeLocaleList.js'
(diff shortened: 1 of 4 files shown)


packages/intl-collator/core.ts
```diff
@@ -84,6 +85,14 @@ export const Collator = function (
   const opt = Object.create(null)
   opt.localeMatcher = matcher
   const collation = GetOption(opts, 'collation', 'string', undefined, undefined)
+  // Resolution option types must be well-formed before negotiation.
+  // https://tc39.es/ecma402/#sec-resolveoptions
+  if (
+    collation !== undefined &&
+    !/^[a-z0-9]{3,8}(-[a-z0-9]{3,8})*$/i.test(collation)
```
longlho: shouldn't this be an abstract method and/or part of Intl.Locale abstract method of sort?

longlho: Agreed. Collator now uses `IsUnicodeLocaleIdentifierType`, added in #7190. Intl.Locale and the other formatters share that helper too. It lives in `ecma402-abstract` to preserve dependency direction. The comment cites ECMA-402 §9.2.8 ResolveOptions, step 6.d.ii, with a pinned source-line link.

PR #7192, opened 2026-09-06
fix(@formatjs/intl-datetimeformat): reject offsets containing seconds

longlho:
Reject time-zone offsets containing seconds, including zero seconds, as required by CreateDateTimeFormat. Hour and minute offsets remain supported and canonicalized.

Closes audit finding 07 from #7185. Updates previous seconds-acceptance tests and adds constructor regressions with a spec-linked code comment.

Validation: `bazel test //packages/ecma402-abstract:ecma402-abstract_test //packages/intl-datetimeformat:intl-datetimeformat_test`.

### DIFF
--- a/packages/ecma402-abstract/IsValidTimeZoneName.ts
+++ b/packages/ecma402-abstract/IsValidTimeZoneName.ts
@@ -1,19 +1,18 @@
 // Cached regex patterns for performance
 const OFFSET_TIMEZONE_PREFIX_REGEX = /^[+-]/
-const OFFSET_TIMEZONE_FORMAT_REGEX =
-  /^([+-])(\d{2})(?::?(\d{2}))?(?::?(\d{2}))?(?:\.(\d{1,9}))?$/
+const OFFSET_TIMEZONE_FORMAT_REGEX = /^([+-])(\d{2})(?::?(\d{2}))?$/
 
 /**
- * IsTimeZoneOffsetString ( offsetString )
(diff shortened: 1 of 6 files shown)


packages/ecma402-abstract/IsValidTimeZoneName.ts
```diff
@@ -27,13 +26,12 @@ function IsTimeZoneOffsetString(offsetString: string): boolean {
     return false
   }
 
-  // 4. Validate component ranges per ECMA-262 grammar
-  // Hour must be 0-23, Minute must be 0-59, Second must be 0-59
+  // CreateDateTimeFormat rejects offsets containing seconds, even zero seconds.
+  // https://tc39.es/ecma402/#sec-createdatetimeformat
```
longlho: need more specific link here showing it's rejected

longlho: [ECMA-402 §11.1.2 CreateDateTimeFormat, step 19.c](https://tc39.es/ecma402/#sec-createdatetimeformat) rejects more than one `MinuteSecond` parse node ([source line 94](https://github.com/tc39/ecma402/blob/b1c961988b9a07894b1dc3dc2b5626ea48387d61/spec/datetimeformat.html#L94)). The second node represents seconds, including `:00`. Moved the citation beside the rejecting match and removed misleading local step numbering. All added spec citations across the stack now include section/step and pinned source lines.

PR #7430, opened 2026-09-16
fix(eslint-plugin-formatjs): generate contracts for nested message helpers

longlho:
When a `defineMessages` catalog mixes plain descriptors with nested `defineMessage` calls, `generateTypes` silently skips the entire catalog. With the default empty ICU contract, parameterized entries then reject their values at call sites.

Resolve nested imported helpers before generating the catalog contract, including aliases and namespace imports. For example, this now generates a numeric contract for `count` and an empty contract for `label`:

```tsx
defineMessages({
  count: {defaultMessage: '{n, number}'},
  label: defineMessage({defaultMessage: 'Terms'}),
})
```

Regression coverage checks mixed catalogs, parameterized nested helpers, and a same-named local function that must not be treated as a FormatJS helper. Public and internal documentation describe the supported pattern.


longlho:
Closing in favor of simplifying the caller: entries inside defineMessages should be plain descriptors rather than redundant nested defineMessage calls. The affected caller has been updated and verified with the published plugin.

